PR #36, opened 2013-05-16
Allow to reverse the order of colours in a brewer palette

jiho:
Use the same argument that hue_pal uses: direction

### DIFF
--- a/R/pal-brewer.r
+++ b/R/pal-brewer.r
@@ -13,7 +13,7 @@
 #' # Can use with gradient_n to create a continous gradient
 #' cols <- brewer_pal("div")(5)
 #' show_col(gradient_n_pal(cols)(seq(0, 1, length = 30)))
-brewer_pal <- function(type = "seq", palette = 1) {
+brewer_pal <- function(type = "seq", palette = 1, direction = 1) {
   pal <- pal_name(palette, type)
 
   # If <3 colors are requested, brewer.pal will return a 3-color palette and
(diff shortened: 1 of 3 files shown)


hadley:
Could you please merge/rebase and add a bullet to NEWS?

jiho:
Done, and squashed everything into one commit.

jiho:
I forgot to write the documentation before. Sorry. `check()` still does not pass on my machine because of problems in creating the PDF documentation. However there is no more info and I don't really know what to fix. Hopefully that's a glitch of my machine and it'll pass on Travis.

jiho:
It passed! It'll be great if you can merge it before the release so that I can depend on it in ggplot2.

R/pal-brewer.r
```diff
@@ -25,9 +28,14 @@ brewer_pal <- function(type = "seq", palette = 1) {
...
-      suppressWarnings(RColorBrewer::brewer.pal(n, pal))[seq_len(n)]
+      pal <- suppressWarnings(RColorBrewer::brewer.pal(n, pal))[seq_len(n)]
     else
-      RColorBrewer::brewer.pal(n, pal)[seq_len(n)]
+      pal <- RColorBrewer::brewer.pal(n, pal)[seq_len(n)]
+
+    if (direction == -1)
+      pal <- rev(pal)
+
+    return(pal)
```
hadley: For future reference, I prefer to use `return()` only for early returns

hadley:
Thanks!

hadley:
BTW if you have a little more time, I'd love a PR with some tests :)

R/pal-brewer.r
jiho: OK. I thought it was good practice to always have an explicit `return()`(it's certainly less confusing for newcomers/students in my experience).

jiho:
What would you want me to test in this particular instance? The order of colors returned? The values of colors returned?

jiho:
Thanks for merging it anyway, I'll get to work on the ggplot end now!

PR #37, opened 2013-11-06
Add function to word wrap long character vectors

jimhester:
This functionality is very useful in plot legends for instance.

Not sure if there is any active development on this package, but I thought I would send the pull request anyway...


hadley:
This seems like a reasonable addition. Can you please:

1. Rebase/merge
2. Add a bullet to news
3. Use a longer name? I think wrap is a bit too general for this specific use

jimhester:
Ok 1c0ea8f should be good to go once the build finishes.

PR #40, opened 2014-02-19
Change dollar to print negative sign on the outside; allow percent to work with negative number

dougmitarotonda:
I am not sure where the authoritative source on this would be, but how about:
http://english.stackexchange.com/questions/124797/how-to-write-negative-currency-in-text

This commit is to fix https://github.com/hadley/scales/issues/38


inst/tests/test-formatter.r
```diff
@@ -53,6 +53,50 @@ test_that("dollar format", {
                  "$600,000", "$700,000", "$800,000", "$900,000", "$1,000,000"))
   expect_equal(dollar(c(9.999)), c("$10"))
   expect_equal(dollar(c(99.999)), c("$100"))
+  expect_equal(dollar(-c(100, 0.23, 1.456565, 2e3)),
```
hadley: These tests seem a bit repetitive. Could you please reduce and break up into separately named tests?

hadley:
I'd be happy to review a changes to add a `use_parens` argument

hadley:
Hmmmm, this got a bit more complicated now that `dollar_format()` is more general - you can supply arbitrary prefixes and suffixes now, and while `-$100` looks better, `-USD 100` certainly doesn't.  Maybe it's just better to implement that parens argument and leave `$-100` as is?

trevorld:
Well character currency descriptors like USD are usually supposed to come after the number, i.e. http://www.proz.com/forum/linguistics/283051-currency_usd_does_it_come_before_or_after_the_number.html so you would expect people to mainly use characters like USD as a suffix which gives you -500 USD in either case (assuming prefix is then "") instead of the more ugly -USD 500...

I think most people (including me and the original patch contributor doug) want negative currency values where the negative sign goes in front of the prefix like -$100 or - £100.45.  Both of us have actually written custom formatters to let us do so since that is the format necessary to have good looking charts if you are an economist or business person.  If for some reason they don't want it reversed then it should be straightforward to write a custom formatter like

my_formatter <- function(prefix, ...) {
   function(x) {
     my_dollar <- dollar_format(prefix=prefix, ...)
     values <- my_dollar(x)
     values <- gsub(paste0("-", prefix), paste0(prefix, "-"), values)
     values
   }
}

PR #46, opened 2014-07-23
Adding a function unit_format

ThierryO:
This function allows users to add units to labels. It has a scale argument that is useful when the data is in another units.

unit_format(unit = "m")(c(1e3, 2e3)) gives c("1,000 m", "2,000 m")
unit_format(unit = "m", sep = "")(c(1e3, 2e3)) gives c("1,000m", "2,000m")
data in m, labels in km
unit_format(unit = "km", scale = 1e-3)(c(1e3, 2e3)) gives c("1 km", "2 km")
data in inch, labels in cm
unit_format(unit = "cm", scale = 2.54)(c(1, 2)) gives  "2.54 cm" "5.08 cm"


hadley:
I wonder if it would be worth allowing something like `unit = "m", scale = "c"` (with a SI unit scale lookup)

ThierryO:
What would the output of unit_format(unit = "m", scale = "c")(c(1e3, 2e3)) look like?

hadley:
I was thinking of it the other way round: `unit_format("m", "k")(c(1e3, 2e3))` would give 1 km, 2 km. (Or with c it would be 100,000 cm, 200,000 cm)

ThierryO:
What if e.g. the data is in mm and you want to display them in cm? I would do unit_format(unit = "cm", scale = 0.1)(c(50, 100)) which returns c("5 cm", "10 cm"). Your suggestion would be to do unit_format(unit = "mm", scale = "c")? I think it would require a lot of parsing.

hadley:
How about:

```{r}
if (is.character(scale)) {
  unit <- paste0(scale, unit)
  scale <- c(m = 1e-3, c = 1e-2, k = 1e3)[scale]
}
```

so you could use either an SI prefix or a multiplier?

ThierryO:
That would work if the value is a basic SI unit (meter, gram, ...). What if the value is not in a basic SI unit (millimeter, kilogram, ...) ?
We might introduce confusion since unit_format(unit = "cm", scale = 10e-2) assumes output is in centimeter and the input is two magnitude higher (in this case meter). One would get the same result with unit_format(unit = "m", scale = "c").

Maybe we can do something like unit_format(input.unit, output.unit, scale). Only two out of the three can be specified.
unit_format(input.unit = "m", scale = "c") meter -> centimeter
unit_format(output.unit = "cm", scale = 1e-2) meter -> centimeter
unit_format(input.unit = "m", output.unit = "cm") meter -> centimeter
unit_format(input.unit = "m", scale = 1e-2) meter -> centimeter
unit_format(input.unit = "hm", scale = "d") hectameter -> decimeter

some tricky things:
- area's unless specified in ares (1 hectare is 100 are, 1 hm² is 10000 m²)
- volumes
- input with non SI units

hadley:
If it wasn't in a SI unit already, you wouldn't use the character argument to scale. But I think you're right that it's too complicated.

hadley:
Can you re-document with latest roxygen, merge/rebase and add a bullet to NEWS?

ThierryO:
I'll do that early next week.

hadley:
Is there any chance you could do it today or early tomorrow? I'd like to submit to CRAN on Friday? 

If you don't have time, just let me know and I can make the changes myself.

ThierryO:
Sorry, I won't have the time to do it before Monday.

hadley:
No problems - now merged!

PR #54, opened 2015-01-15
Removed stray period

seaaan:
Alternative punctuation: "Consider specifying shapes manually. If you must have them."


hadley:
Thanks!

PR #56, opened 2015-05-24
Fix truncated documentation in hue_pal chroma

zeehio:
Documentation in `hue_pal` for the chroma parameter was truncated because of a `#` instead of a  `#'`. This affects `scale_color_discrete` and others from ggplot2, that get the documentation from here too.

This pull request has two commits: 
- A commit with the one-line fix
- A commit rebuilding the documentation.

Just in case you need this statement: no copyright applies to the patch as it is trivial.

Thanks for your time and congratulations for your work!


hadley:
Thanks!

PR #57, opened 2015-05-26
Color mapping features from leaflet and rasterfaster

jcheng5:
Adds `col_numeric`, `col_bin`, `col_quantile`, `col_factor`, and `colour_ramp` functions.


jcheng5:
Note that this brings in Rcpp and RcppParallel, the latter of which causes these kinds of warnings: https://github.com/RcppCore/RcppParallel/issues/5

I can easily switch to non-parallel if you prefer, and drop the RcppParallel requirement. The parallel is faster but not markedly so (maybe 2X I think).

jcheng5:
Oh, and camel case to underscore. OK, I have some work to do...

R/colour-mapping.r
```diff
@@ -0,0 +1,386 @@
...
+#'   domain.
+#' @param na.color The color to return for \code{NA} values. Note that
+#'   \code{na.color=NA} is valid.
+#'
+#' @return A function that takes a single parameter \code{x}; when called with a
+#'   vector of numbers (except for \code{col_factor}, which expects
+#'   factors/characters), #RRGGBB color strings are returned.
+#'
+#' @export
+col_numeric = function(palette, domain, na.color = "#808080") {
```
hadley: Can you use `<-` for assignment please?

jcheng5: Done.

jcheng5:
@hadley Do you mind taking a look at this when you get a chance? We want to take Leaflet to CRAN soon and we need to decide whether we want a dependency on scales or rasterfaster.

Thanks!

hadley:
LGTM.  Do you want me to plan a scales release?

jcheng5:
That would be great, thanks--I can basically release Leaflet once scales is
up on CRAN.

Issue #43, opened 2014-06-18
Request: Additional argument to percent()


aviast:
The Commonwealth Style Manual requires that percentages be followed by " per cent", not a % sign.

Request that percent be modified to support this.
eg.
percent(x, sign = "%") { ... }

hadley:
I'd be happy to review a PR

PR #60, opened 2015-06-14
Modified percent_format to accept user-specified sign

aviast:
Modified percent_format to accept user-specified sign as per issue #43


hadley:
Looks good - could you please merge/rebase and add a bullet point to NEWS?

PR #70, opened 2016-02-12
Go back to a more strict numeric check in train_continuous

cpsievert:


hadley:
That fails for dates/date times so definitely isn't what we want

cpsievert:
I agree. I submitted this only to demonstrate its effect on the example in https://github.com/hadley/ggplot2/issues/1542

PR #71, opened 2016-02-12
Set R_LIBS_SITE

jimhester:
Trying to debug travis failures with r_binrary packages


hadley:
I was thinking maybe we should have a set of repos with different configs and a script that uses the travis api to restart them all when you want to make changes

jimhester:
Yeah that seems like a decent idea

Issue #74, opened 2016-03-01
rescale Date (and POSIXct) objects?


zeehio:
Hi, 

As of today it is not possible to rescale dates (and times) using `scales::rescale`. Is there a particular reason? I am puzzled by this, and I think that I must be missing something, because the problem seems fairly simple to tackle (using `as.numeric` somewhere). I would like to contribute with a clean solution, but I am afraid I am being naive on the solution. Any advice? (ggplot2::geom_segment [could benefit](https://github.com/hadley/ggplot2/issues/1048) from this)

```
days <- as.Date(c("2016-03-01", "2016-03-02", "2016-03-03"))
scales::rescale(days[2], from = c(days[1], days[3])) == 0.5

datetimes <- as.POSIXct(c("2016-03-01 08:05:00",
                          "2016-03-01 08:10:00",
                          "2016-03-01 08:15:00"),
                        format = "%Y-%m-%d %H:%M:%S")
testthat::expect_equal(scales::rescale(datetimes[2],
                       from = c(datetimes[1], datetimes[3])), 0.5)
```

I get

```
Error in Math.Date(x) : abs not defined for "Date" objects
```

hadley:
I think coercing back and forth between numerics is probably the right thing to do.

zeehio:
I just rebased the pull request on the latest master. Is there anything else I can do to get this merged?

Thank you for your time

zeehio:
This is already merged in #75, so I will close it

PR #75, opened 2016-03-01
rescale() works with Date, POSIXct and POSIXlt objects.

zeehio:
This pull request deals with #74 as discussed.

Changes:
- documents the change in `NEWS.md`
- adds tests for rescaling dates (simple tests but enough to show and test the feature).
- edits `R/bounds.r`, checking for `Date`, `POSIXct` and `POSIXlt` inputs; converting them to numeric.

Thanks for your time. Feel free to tell me if there is anything that does not convince you.


R/bounds.r
```diff
@@ -1,3 +1,13 @@
+#' If needed, converts input from classes that lack division and multiplication
+#' to classes that can be scaled.
+#' @param x vector of values to convert, if needed
+adapt_input_class <- function(x) {
```
hadley: This shouldn't be documented

R/bounds.r
```diff
@@ -1,3 +1,13 @@
...
+  if (any(class(x) %in% c("Date", "POSIXct", "POSIXlt", "POSIXt"))) {
```
hadley: Please use `inherits()` here.

A better formulation would be:

```R
if (...) {
  ...
} else {
 ...
}
```

as then you no longer need explicit `return()` statements..

zeehio: Done, thank you for your quick review.

karawoo:
This would be useful for some of the scales stuff I'm working on for ggplot2 – it would allow scaling alpha with datetimes, for one. But I think we need a way to go back to datetimes after the conversion, otherwise you get legends that look like this:

```r

dat <- data.frame(
  x = sample(1:10, 10),
  y = (1:10)^2,
  dt = seq(
    as.POSIXct("2017-06-01 15:28:44 PDT"),
    as.POSIXct("2017-06-12 15:28:44 PDT"),
    length.out = 10
  )
)

ggplot(dat, aes(x, y)) +
  geom_point(aes(alpha = dt))
```

scale_alpha_datetimes

zeehio:
@karawoo Thanks for reviewing this. I believe that what you are asking for fits better in the `ggplot2` package and not in the scales package.

Here is your example, with a draft of what the `ggplot2::scale_alpha_datetime` could be:

```
library(ggplot2)
dat <- data.frame(
  x = sample(1:10, 10),
  y = (1:10)^2,
  dt = seq(
    as.POSIXct("2017-06-01 15:28:44 PDT"),
    as.POSIXct("2017-06-12 15:28:44 PDT"),
    length.out = 10
  )
)


scale_alpha_datetime <- function(name = waiver(), origin = "1970-01-01 00:00.00 UTC", ...) {
  origin <- force(origin)
  label_fun <- function(breaks) {
    as.POSIXct(breaks, origin = origin)
  }
  scale_alpha_continuous(name = name, labels = label_fun, ...)
}


ggplot(dat, aes(x, y)) +
  geom_point(aes(alpha = dt)) +
  scale_alpha_datetime()
```

imatge

The actual `scale_alpha_datetime` function should be closer to the other `scale_*_datetime` functions, to provide better label formatting and a more reasonable choice of date breaks.

R/bounds.r
```diff
@@ -1,3 +1,14 @@
+# If needed, converts input from classes that lack division and multiplication
+# to classes that can be scaled.
+# @param x vector of values to convert, if needed
+adapt_input_class <- function(x) {
```
hadley: I'm not sure I follow the reasoning here. Why not just unconditionally coerce to numeric?

zeehio: `scales::rescale` mainly depends on elementary arithmetic operations (addition, subtraction, multiplication and division). Coercing to numeric is a good way to define multiplication and division operations for `POSIXct` objects, but it is not the way to go for all R objects. For instance `integer64` numbers, from the `bit64` package, define their own arithmetic operators to prevent loss of precision.

If the coercion to numeric is done, there will be a warning when using some integer64 objects:

```
x <- bit64::as.integer64(c(2^50, 2^54+1, 2^60+2))
from <- range(x, na.rm = TRUE)
# I need to compute the `from` range here due to an unrelated bit64 bug triggered
# when range(..., finite = TRUE) is used. (this bug has already been
# reported to the bit64 maintainer)

# No warning given now:
scales::rescale(x = x, from = from)
# [1] 0.00000000 0.01466276 1.00000000

# If we force coercion to numeric there is a warning:
scales::rescale(x = as.numeric(x), from = as.numeric(from))
# Warning messages:
# 1: In as.double.integer64(from) :
#   integer precision lost while converting to double
# 2: In as.double.integer64(x) :
#   integer precision lost while converting to double
# [1] 0.00000000 0.01466276 1.00000000
```

I wanted to provide an example were the `as.numeric` approach is wrong, but `scales::rescale` calls `scales::zero_range` that has a default tolerance too large for `integer64` data. I could work towards letting the user give a custom tolerance value, but for the sake of the explanation you can set `tol` to zero manually in the `zero_range` definition in `r/bounds.r` and rebuild, in case you want to reproduce this example:

```
x <- bit64::as.integer64(2^60) + 1:3 
# x has three consecutive numbers, we expect the result to be c(0, 0.5, 1)
x_range <- range(x, na.rm = TRUE)
scales::rescale(x = x, from = x_range) # Correct result
# [1] 0.0 0.5 1.0
scales::rescale(x = as.numeric(x), from = as.numeric(x_range)) # loss of precision
# [1] 0.5 0.5 0.5
```

Even if we have tolerance issues in this case, I prefer to work towards making `scales` work with other objects, and therefore not coercing everything to `numeric`.

But you have the last word, though :smiley:

hadley: In that case I'd prefer to use an S3 generic.

zeehio: Using an S3 generic is the best approach, my apologies for not thinking about them when I wrote this patch. I was now able to provide a method (and test case) for `integer64` data as well.

hadley:
Thanks!

PR #77, opened 2016-05-05
More options for Percent format

larmarange:
The purpose of this pull request is to:
- allow the user to control the rounding
- allow to display accurate percentage in other languages than English (requires to be able to specify big.mark, decimal.mark, prefix and suffix)
- to display per mille or even parts per million (ppm) values


larmarange:
NB: solve https://github.com/hadley/scales/issues/72

larmarange:
I'm still interested by this PR and I will update code style and add some unit tests.

Just one remark. This updated version of `percent_format` is quite generic and I was thinking if it would be relevant or not to provide a generic `number_format(accuracy = 1, multiplier = 1, prefix = "", suffix = "", big.mark = " ", decimal.mark = ".", scientific = FALSE, ...)` function and then to use it to generate others specific functions just by changing the default values, e.g:

```r
comma <- function(x, ...) {number_format(big.mark = ",", ...)}
percent <- function(x, ...) {number_format(multiplier = 100, suffix = "%", big.mark = ",", ...)}
unit <- function(x, unit = "m", scale = 1, sep = " ", ...) {
  number_format(multiplier = 1 / scale, suffix = paste0(sep, unit), ...)
}
```

If you do not think it is appropriate, I will just update `percent_format`.

hadley:
Yes, I think a generic numeric formatter that could power all of these functions would be useful.

larmarange:
So following the discussion, I have added a generic `number_format` allowing to specify scale, suffix, prefix, accuracy...

`comma_format`, `unit_format` and `percent_format` are now specific cases of `number_format` with just different default values.

`ordinal_format` and `dollar_format` also rely on `number_format`, but with additional actions and backward compatibility. For `ordinal_format`, it is also possible to define custom rules for ordinal indicators, allowing to adapt the function for other languages than English.

All these functions now share a consistent set of parameters.

NB: `NEWS.md` has not been updated.

Let me know if you have any comment

hadley:
Let's come back to this after the other PRs have been dealt with. I think it's a worthwhile change but there are probably going to be some merge conflicts.  We'll probably also need to rethink the docs

dpseidel:
Going through issues related to or potentially affected by this PR -  I noticed that the bug in #97, namely that `percent()` returns all `NaN`s when any element in the passed vector is `Inf`, will now throw a fairly obscure error (see below). It's perhaps worth considering explicit handling of `Inf` values or simply an improved error message.

Current scales 0.5.0 behavior:
```
library(scales)
percent(c(.1, .3, Inf))
#> [1] "NaN%" "NaN%" "NaN%"
```

Behavior given this PR
```r
percent(c(.1, .3, Inf))
# Error in prettyNum(.Internal(format(x, trim, digits, nsmall, width, 3L,  : 
#   invalid 'nsmall' argument
# In addition: Warning messages:
# 1: In min(x) : no non-missing arguments to min; returning Inf
# 2: In max(x) : no non-missing arguments to max; returning -Inf
# 3: In precision(x) :
```

larmarange:
I have added some code to deal properly with Infinite values, as expressed in your earlier comments.

For now, I have not solve the merging conflicts, as some other merge conflicts are still possible.

hadley:
To avoid the merge conflicts, how about we tackle this in two phases? 

1. Create the generic number formatter - it can go in its own file, and hence have tests in a separate file.
2. Use the generic formatter in `percent_formatter()` etc

If you think that's a good strategy, can you please close this PR and start a new one?

larmarange:
You mean a new PR without applying Roxygen? (i.e. with just the changes in R code and in test file)

Do you expect other changes in formatter.R? Otherwise, it would be easy to solve conflicts and to rebase the PR into one commit.

hadley:
Regardless of the merge conflicts, this PR is rather large, so I think it would be easier to review in two parts: one that provides the general numeric formatter (including docs and tests) and then one that adapts the existing functions to use it.

larmarange:
As proposed, I have prepared a new PR with only the new generic number formatter: https://github.com/r-lib/scales/pull/142

One the generic formatter with all its arguments validated, we will have to update the existing formatters to rely on the generic one.

PR #78, opened 2016-09-11
Including the option to set a different decimal mark. 

flaviobarros:
Can be useful in countries where "," is used instead of ".", like Brazil. I have been using this new feature to write apps with brazilian reals, where we use "," as the decimal character.

Every time that the user try to set just big.mark to "." it raises a warning advising the user to set a different decimal mark.


flaviobarros:
I need to fix documentation before.

PR #86, opened 2017-05-04
added label size control to show_col()

AndreaCirilloAC:
problem: trying to plot great enough matrices ( 400X400 on...) produces unreadable labels
proposed solution: giving the user control over label size.

notes:
- parameter named cex_text consistently with plot() function
- updated show_col() function and related documentation


AndreaCirilloAC:
than you Travis, I found out I didn't update the example.
going to fix it an try again.

PR #90, opened 2017-06-22
Add viridis palette

karawoo:
Adds the viridis palette originally from [ggplot2 PR](https://github.com/tidyverse/ggplot2/pull/2178).


hadley:
Directly in master, can you please do `use_coverage()` to get rid of the annoying codecov notification?

karawoo:
Added that and deleted the codecov comment. I can also add some tests if that would be useful, I wasn't sure what behavior we'd want to test.

PR #91, opened 2017-06-26
Calculate minor breaks as a property of the transformation

karawoo:
This adds a `calculate_minor_breaks()` function to calculate minor breaks based on existing major breaks and limits. The `reverse` argument indicates whether the minor breaks should get calculated according to a reversed scale or not; this defaults to `FALSE` but is set to `TRUE` for `reverse_trans()` (which addresses ggplot2 issue [2081](https://github.com/tidyverse/ggplot2/issues/2081)). 

After this change [these lines](https://github.com/tidyverse/ggplot2/blob/04a707ed6cd3b3e19a07bd5226a2fa44a8e8be91/R/scale-.r#L275-L282) in ggplot can be replaced with:

```r
breaks <- self$trans$minor_breaks(b, limits, n)
```


R/breaks.r
```diff
@@ -150,3 +150,38 @@ cbreaks <- function(range, breaks = extended_breaks(), labels = scientific_forma
...
+#'
+#' @param reverse if TRUE, calculates the minor breaks for a reversed scale
+#' @export
+#' @examples
+#' m <- extended_breaks()(1:10)
+#' calculate_minor_breaks()(m, 1:10, n = 2)
+#' 
+#' n <- extended_breaks()(0:-9)
+#' calculate_minor_breaks(reverse = TRUE)(n, 0:-9, n = 2)
+calculate_minor_breaks <- function(reverse = FALSE) {
```
hadley: I think just call this `minor_breaks()`

R/breaks.r
```diff
@@ -150,3 +150,38 @@ cbreaks <- function(range, breaks = extended_breaks(), labels = scientific_forma
...
+      if (min(limits) < min(b)) b <- c(b[1] - bd, b)
+      if (max(limits) > max(b)) b <- c(b, b[length(b)] + bd)
+    } else {
+      if (max(limits) > max(b)) b <- c(b[1] - bd, b)
+      if (min(limits) < min(b)) b <- c(b, b[length(b)] + bd)
+    }
+    
+    breaks <- unique(
+      unlist(
+        mapply(seq, b[-length(b)], b[-1], length.out = n + 1, SIMPLIFY = FALSE)
```
hadley: Slightly cleaner to use `Map()` instead of `mapply()` + `SIMPLIFY = FALSE`

And maybe make a `seq_between` helper?

```R
seq_between <- function(a, b) {
  seq(a, b, length.out = n + 1)[-(n + 1)]
}
```

I think that makes the code slightly clearer.  (That might also enable you to eliminate the `unique()`, I'm not sure)

R/trans-numeric.r
```diff
@@ -111,7 +111,10 @@ reciprocal_trans <- function() {
 #'
 #' @export
 reverse_trans <- function() {
-  trans_new("reverse", function(x) -x, function(x) -x)
+  trans_new("reverse",
```
hadley: Can you use tidyverse indenting here please?

R/breaks.r
```diff
@@ -150,3 +150,38 @@ cbreaks <- function(range, breaks = extended_breaks(), labels = scientific_forma
...
+#' n <- extended_breaks()(0:-9)
+#' calculate_minor_breaks(reverse = TRUE)(n, 0:-9, n = 2)
+calculate_minor_breaks <- function(reverse = FALSE) {
+  function(b, limits, n) {
+    b <- b[!is.na(b)]
+    if (length(b) < 2) return()
+
+    bd <- diff(b)[1]
+
+    if (!reverse) {
```
hadley: And maybe add a comment here? `# Allow minor breaks to extend outside breaks towards limits`

R/breaks.r
karawoo: That causes a recursive default argument error unless I change the name of the new `trans_new()` argument. I could change that argument to `minor.breaks`?

R/breaks.r
karawoo: It does remove the need for `unique()` but it fails to return the final break.

R/breaks.r
hadley: Maybe we should name more like the other breaks functions, to also include the strategy. That would make it `even_minor_breaks()`

R/breaks.r
hadley: Oooh, the final final break. You could add that back by hand?

karawoo:
@hadley I think I addressed all your comments, let me know if this looks ok to you

R/breaks.r
```diff
@@ -150,3 +150,42 @@ cbreaks <- function(range, breaks = extended_breaks(), labels = scientific_forma
 
   list(breaks = breaks, labels = labels)
 }
+
+#' Minor breaks.
```
hadley: Titles shouldn't end in a full stop (I've inconsistently applied this rule in the past, but I've been trying to do better going forward)

R/breaks.r
```diff
@@ -150,3 +150,42 @@ cbreaks <- function(range, breaks = extended_breaks(), labels = scientific_forma
...
+#'
+#' @param reverse if TRUE, calculates the minor breaks for a reversed scale
+#' @export
+#' @examples
+#' m <- extended_breaks()(c(1, 10))
+#' even_minor_breaks()(m, c(1, 10), n = 2)
+#' 
+#' n <- extended_breaks()(c(0, -9))
+#' even_minor_breaks(reverse = TRUE)(n, c(0, -9), n = 2)
+even_minor_breaks <- function(reverse = FALSE) {
```
hadley: Maybe `regular` would be better than `even`?

karawoo: Yeah I like that better too

PR #92, opened 2017-06-26
Add additional methods to `rescale` and `rescale_mid`

karawoo:
#75 didn't include a `rescale.NULL` method, which is sometimes needed. For example this:

``` r
library("ggplot2")
ggplot(mtcars, aes(cyl, mpg)) +
  geom_point() +
  scale_x_continuous(breaks = NULL)
```

should create a plot like: 

But currently it throws `Error in UseMethod("rescale"): no applicable method for 'rescale' applied to an object of class "NULL"` instead. The lack of `rescale.NULL` also broke some code in the `ggplot2-specs` vignette.


zeehio:
Thanks @karawoo. Do you think `scales` needs to provide a `rescale_mid.NULL` method as well?

hadley:
Looks good. As @zeehio suggests, can you please also add `rescale_mid.NULL`?

karawoo:
Will do. We also need `logical` methods for [this test](https://github.com/tidyverse/ggplot2/blob/04a707ed6cd3b3e19a07bd5226a2fa44a8e8be91/tests/testthat/test-scales.r#L39) to pass.

PR #104, opened 2017-10-17
Fix `rescale_mid()` to properly handle NAs.

foo-bar-baz-qux:
Noticed in the course of fixing one of the ggplot2 bugs that `rescale_mid` handles `NA`s differently than the other rescale functions. This updates `rescale_mid` to be inline with those other functions when handling `NA`s.


hadley:
Could you please add a unit test?

hadley:
@dpseidel can you please finish this one off by adding a test?

tests/testthat/test-bounds.r
```diff
@@ -9,8 +9,9 @@ test_that("rescale_mid returns correct results", {
 
   expect_equal(rescale_mid(x, mid = 1, to = c(0, 10)), c(0, 2.5, 5))
   expect_equal(rescale_mid(x, mid = 1, to = c(8, 10)), c(8, 8.5, 9))
-})
 
+  expect_equal(rescale_mid(c(1, NA, 1)), c(0.5, NA, 0.5))
```
hadley: I think this ok here, but it would be better to create a new test that fully exercises missing values (i.e. also checking what happens when the range isn't 0), since there are no existing tests for NAs.

dpseidel: Agreed. Full tests of NA handling added in 7f7bdd5.

Issue #102, opened 2017-10-03
rescale stopped working with dist objects


Bisaloo:
`rescale()` used to work with `dist` objects as exemplified [here](https://onlinelibrarystatic.wiley.com/store/10.1111/2041-210X.12735/asset/supinfo/mee312735-sup-0004-AppendixS4.R?v=1&s=7eecca9a563372b23be43debb5f1d2aafcaba168) (SI from https://doi.org/10.1111/2041-210X.12735).

It was broken by #75. 

> Error in UseMethod("rescale") : 
>  no applicable method for 'rescale' applied to an object of class "dist"

Do you explicitly want to remove support for `dist` objects or is it simply a side-effect of the changes in #75?

zeehio:
Hi @Bisaloo, as the author of #75 I am sorry for the inconveniences. I wish I had seen your issue earlier. I have already submitted a pull request, would you like to try the patch?

```
devtools::install_github("zeehio/scales@fix_102")
```

Bisaloo:
It's exactly the old function so I expect everything will be fine. I just tested on a random `dist` object just in case and it works fine.

Thank you for taking care of this!

PR #105, opened 2017-10-17
Provides a default method for rescale and rescale_mid

zeehio:
In https://github.com/hadley/scales/issues/74 and https://github.com/hadley/scales/issues/75 we used a generic S3 method for `rescale` and `rescale_mid`. The aim was to handle more types of objects (POSIXt objects...) and to prevent errors like the loss of precision that happened when coercing `integer64` to `numeric`.

While that PR fixed the support for POSIXt types, it also brought issue #102. 

There are many numeric types out there that do not inherit from the numeric
class (like the `dist` object mentioned in #102), compared to the number of cases similar to integer64, where the numeric coercion is wrong.

So there is a trade-off:

- Not providing the `default` method in `rescale` and `rescale_mid` has the advantage that we only rescale the classes that we know can be rescaled, and the disadvantage that we get issues like #102 when we find a new object. This approach is the current one and it was implemented in https://github.com/hadley/scales/issues/75

- Providing `rescale.default` and `rescale_mid.default` has the advantage of catching all the numeric-like types that do not inherit the numeric class (such as the dist object) and working as expected. However, we risk to deliver wrong results without an error if a case similar to integer64 appeared. This is what this pull request follows.

The changes of getting more issues like #102 are much higher than getting issues like the loss of precision in integer64 objects, so in my opinion this pull request makes sense.


dpseidel:
Hi @zeehio

We definitely want to address #102 but prefer to be stricter and implement a single new method for `dist` objects instead of opting for the `default` method approach you suggest here.  

Are you interested in making the necessary changes?

zeehio:
Hi @dpseidel 

Thanks for your time working on the scales package and for reviewing this pull requests. Here is the support for `dist` objects as requested. I reverted my first commit to keep the other solution in the git history, in case we ever need to go back to it. I also rebased the pull request.

Please tell me if there is anything else to be done on my end :smiley:

PR #106, opened 2017-10-25
Pseudo-log transform

lepennec:
Hi,

this is a simple implementation of the pseudo-log transform, a.k.a as the asinh transform. This is a transform which behaves as sign(x)*log(|x|) when x is large and as x when x is small. Such a transform was mentionned in the issue https://github.com/hadley/scales/issues/48. I've added two parameters, one which is used to define when x is small or not and the other to mimic a logarithm with a different base than the natural one (exp(1)).

Erwan


lepennec:
Yes. Let me know how I can help.

dpseidel:
Hi @lenpennec, 

We need a couple small things before this is ready for merge: 

- [ ] fix styling. Tidy code styling is easily accomplished with the package `styler`. In this case, the only thing that jumps out at me is that the function documentation should have a line between the title and the params for readability/consistent roxygen style.
- [ ] can you add a test for this transformation? 
- [ ] can you add a NEWS bullet explaining the change? 

Thanks! Let me know if I can help.

dpseidel:
@lepennec, 

One final thing: the pattern for test file naming is to match the test file name to the R file name (as `usethis::use_test()` does). So rather than "test-psuedo-log", the appropriate file name is "test-trans-numeric" and the `context` similarly matched, something like  "Trans - numeric" .  Thanks so much for your contribution and patience with this PR!

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # scales 0.5.0.9000
 
+* New `pseudo_log_trans()` for transforming numerics into a signed logarithmic scale
```
hadley: Needs username

hadley:
@dpseidel since the last change is small, can you please make it and merge?

PR #109, opened 2017-12-20
Respect alpha by default in `colour_ramp()`. Closes #108

### LINKED ISSUE #108: Function colour_ramp() should respect transparency by default
The function `colour_ramp()` has the option to ramp RGBA colors instead of RGB colors. However, by default, this option is switched off, and the function removes any alpha values that may be attached to the colors it is given.

This seems to be an unusual default setting. If somebody goes to the effort of adding alpha values to colors, I would assume they want those respected. Also, this creates inconsistency between discrete and continuous color scales. Discrete scales respect alpha value, continuous ones do not.

Reproducible example: First a case using a discrete color scale.

```
library(ggplot2)

ggplot(iris, aes(Sepal.Length, fill = Species)) + 
  geom_density() +
  scale_fill_manual(values = c("#00FF00FF", "#008000B0", "#0000FF20"))
```

screen shot 2017-12-19 at 6 59 33 pm

The transparency values of the colors are respected. Now the same plot using a continuous color scale:
```
ggplot(iris, aes(Sepal.Length, fill = as.numeric(Species), group = Species)) + 
  geom_density() +
  scale_fill_gradientn(colors = c("#00FF00FF", "#008000B0", "#0000FF20"))
```

screen shot 2017-12-19 at 6 59 45 pm

Transparency is gone.

However, if we switch the `alpha` option to `TRUE` in function `colour_ramp()`, things behave as expected:

```
# change the default setting of scales::colour_ramp:
assignInNamespace("colour_ramp", function(colors, na.color = NA, alpha = TRUE){
  if (length(colors) == 0) {
    stop("Must provide at least one color to create a color ramp")
  }
  colorMatrix <- grDevices::col2rgb(colors, alpha = alpha)
  structure(function(x) {
    scales:::doColorRamp(colorMatrix, x, alpha, ifelse(is.na(na.color), 
                                                       "", na.color))
  }, safe_palette_func = TRUE)
}, "scales")

ggplot(iris, aes(Sepal.Length, fill = as.numeric(Species), group = Species)) + 
  geom_density() +
  scale_fill_gradientn(colors = c("#00FF00FF", "#008000B0", "#0000FF20"))
```

screen shot 2017-12-19 at 7 00 09 pm

clauswilke:
See #108 for an explanation why I think the default setting should be changed.

One comment: As `colour_ramp()` is currently implemented, with this setting it always converts RGB colors to RGBA. I could add some checks that make the function return RGBA only if at least one RGBA color was provided as input. Not sure whether you'd prefer that behavior over consistently returning RGBA values.


clauswilke:
The unit tests are failing because RGBA colors are returned where RGB colors were expected. Before addressing this, I will wait for input on whether I should change the expected output in the unit tests or change `colour_ramp()` so it doesn't return RGBA unless provided with RGBA input.

hadley:
I always worry about unintended consequences of changes to stuff deep in the internals of ggplot2 that haven't been touched in a while and are unlikely to be well tested. It would be useful if you could also run the vdiffr visual tests for ggplot2 and see if anything changes (although presumably this will only affect gradient scales?)

I think always returning RGBA is likely to fine. (Although it might cause test failures in downstream packages, so checking for an alpha component might be safer)

clauswilke:
It would be very easy to have a check in the C++ function to only add an alpha component if there's actually some non-zero transparency. If we do that on a color-by-color basis, though, we might end up with cases where a vector contains a mix of RGB and RGBA colors. Can you think of a scenario where that might cause a problem?

I'll run the visual tests for ggplot2.

hadley:
That seems ok to me - it's not changing the colour, it's just ensuring that the new behaviour matches the previous as much as possible so we don't spuriously break tests. I don't think you even need the `compact_rgb` option

clauswilke:
I ran the visual tests for the current development branch of ggplot2. Most tests pass, and the ones that fail all contain a continuous color scale.

```
> manage_cases()
Running testthat to collect visual cases

  N = New visual case
  X = Failed doppelganger
  o = Convincing doppelganger

.............................................oooooo.o..................o...ooo.......oooooXoooooo......................................................oo....................o................................................................................................................................................WWW.......................o........ooooooooooooooooooooooooooooooo.o.......ooooo....ooooo.XXXXXXXXXXXo...........oo.............oooooooooooooo.....F............................oooXXXXXXXXXXXXXXXXX............................................................................................................................................................................oooooooo............................................................................................................................ooooooooooooo......................................................o..........................................................................ooo........................................ <truncated>
```
I tracked down the origin of the failing comparisons in the svg files. For all cases I checked, it was the image that represents the legend bar:

```$diff stat-density2d-with-filled-polygons.svg stat-density2d-with-filled-polygons_old.svg 
64c64
< <g clip-path='url(#cpMHw3MjB8NTc2fDA=)'><image width='17.28' height='86.40' x='670.87' y='246.24' xlink:href='data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABIAAABXCAYAAAAJWxWIAAAABmJLR0QA/wD/AP+gvaeTAAABMElEQVRYhc2YwXUDIAxDJcJyXaArdL7skl1yAakH2gn4hzCAnyWEZOPv57u6PJW/xm2R/zOb3FdpmUK2Ndt9XWjbmiKgpZrNfUenUIGOZM0A0GoKmkqRbYYjC4K2DSlbMqUjQWRTHLmDubUoENkDKuRIU8SjpXS0BTmkTJp/AB3VgcgOBY26NddUQEKPdqmUsiFoZ4gAlP1n/oAfhYL2cWRjAelAZC8sIEt5NqUjU2NNqI5AHWEBOSirDWMjS1hAUtCwgKRyLZSyuaSlbMTDENmLckhTZG+WbGCq5TybWiEiaM6WqJ020OJXqCMujuC9HxCkqdfP7SJQiugE5AeRnVJki/pjw+ajs9Rc//lqS5qixmPmH5Lyo0F1RCWtqaRNqing+iWJ+/O3/HNbpI/1+gVxyUggzVd1wgAAAABJRU5ErkJggg=='/></g>
---
> <g clip-path='url(#cpMHw3MjB8NTc2fDA=)'><image width='17.28' height='86.40' x='670.87' y='246.24' xlink:href='data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABIAAABXCAYAAAAJWxWIAAAABmJLR0QA/wD/AP+gvaeTAAABNklEQVRYhc2YwVkGQQxCYUxzNmAL1mcv9uJlBv6DawXzDm4B+RYCJBN/fP1Ul1/l93Vb5O+b9vqHJB9Nc+7ryJom14XOElNIiUYFClkMRypUyK6mALTEEDRD7XcL6cjWqAC0QNBkCJoKdc1U1yIK2qKgRZT7KUHKmhKCNJWQlNdoHQEcLSghvalgczUioAmCtqmuPTqC5hpi2oMKEjFtIB0Jcr9D6UjW6F8NSCpquWArRLaoPdvoMspMWgjaDjYgFwRNlPsXlEfZi9pqsWCjRrahHXJXVLAFcv+hwp8i+/eZhWQ2Ff4xFbUU2ZRFIvZRQ5w0FmRaKiFPqXEkSkfsux8hG1J2ujQKcYfcFNnUfkStx88higg26g5JnQ8fsu/bb+5WC1kkqUbEhV0Sd/O3/HlbpG/7+wX6zEQzSzM5XAAAAABJRU5ErkJggg=='/></g>
```
I compared the legend bar images visually and didn't see any difference. They look the same to me. 

screen shot 2017-12-20 at 2 55 12 pm

hadley:
It's probably some tiny difference in the way colours are recorded that makes the embedded png slightly different. It seems fine to me. 

But because there are changes to the visual tests, we'll also need a parallel PR to ggplot2 that bumps the required scales version and saves the updated tests.

clauswilke:
I think the failing tests are unrelated to my changes in scales. I did another test that I should have done earlier. I reverted to the current scales package on CRAN and re-ran the visual tests. I got exactly the same failing tests, with the same differences. The two svgs created on my machine, for scales 0.5.0 and 0.5.0.9000, are exactly the same.

```
wilke$ diff rays-circular-arcs-and-spiral-arcs_scales0.5.0.svg rays-circular-arcs-and-spiral-arcs_scales0.5.0.9000.svg
wilke$ 
```

So it's probably the old problem of visual tests not being reliable across different machines. In fact, I don't expect there to be any difference caused by my change to scales, because there shouldn't be any unless somebody specified an RGBA instead of an RGB color as input to `colour_ramp()`, and nowhere in the ggplot2 code does that happen as far as I know.

clauswilke:
Hadley, I was wondering: Since all the visual tests for ggplot2 need to be redone anyways, would it be possible to incorporate this patch before that happens?

dpseidel:
Hi @clauswilke

After a few final changes, I think we'll be ready to merge: 
- [ ] the tests need styling to conform tidy style (i.e. line length, spacing, etc). `styler` can help with this.
- [ ] test arguments should be switched to reflect the (`object`, `expected`, ...) structure the `expect_equal` function and you can remove the `scales::` namespace calls. 
- [ ] Given the detailed comments added between tests, you may want to consider breaking these up into individual more-specific expectation-test chunks.
- [ ] Finally, it would be nice to update the PR fork and resolve the merge conflict locally 

Let me know if there is anything you'd like me to do on my end.

clauswilke:
@dpseidel I'll try to get these things done this weekend.

clauswilke:
I have rebased and made the requested code changes in the testing code. I have also verified that all ggplot2 regression tests continue to work.

One other change I see is that the documentation has now been generated with the development version of roxygen. Please let me know if this is a problem. I believe that roxygen version is now required for ggplot2 development, I wouldn't know why else I'd have it.

Issue #100, opened 2017-09-17
log-scale Breaks Are Not Always Useful


billdenney:
This is tidyverse/ggplot2#2269 and somewhat related to #9.

When plotting with ggplot2 and using log-scale axes, breaks don't always show up.  This is related to the fact that log_breaks doesn't always provide breaks within the range.  A specific example would be:

```r
> scales::log_breaks()(2:3)
[1]  1 10
```

When a plot is generated, both 1 and 10 are often outside of the range, and the number of breaks requested is 5 when only 2 are returned which also appears to be a bug.  More useful (in most scenarios that I think of) would be to select human-readable numbers that are approximately within the range of interest.

Specifically, I think that the preferred scales (at least for log 10) would minimize the number of decimal significant figures while matching the requested number of breaks with at least `n - 2` of the breaks within the range of interest. (`n - 2` rather than `n` to allow for the fact that being just outside the range of interest could still show up on the scale.)  The right metric on scales other than log10 may be to minimize the number of significant figures in the exponent.

For example, the following still doesn't match the requested number of breaks being 5 nor are the breaks within the range.

```r
> scales::log_breaks(base=exp(1))(c(2, 2.0001))
[1] 1.000000 2.718282
```

This function appears to better align with the hope of showing log-scale breaks even when the range does not align with the desire of generating breaks for plotting:  The number of requested breaks are returned, and the breaks are (approximately) equally spaced on the log scale:

```r
log_breaks <- function(n=5, base=10) {
  function(x) {
    rng <- log(range(x, na.rm = TRUE), base = base)
    min <- floor(rng[1])
    max <- ceiling(rng[2])
    
    if (max == min) {
      ret <- base ^ min
    } else if ((max - min) <= (n - 2)) {
      # Approximately minimize the number of log10 significant figures
      ret <- seq(rng[1], rng[2], length.out=n)
      rounding_digits <- -floor(log10(diff(base^ret)))
      rounding_digits <- c(rounding_digits[1], rounding_digits)
      ret <- round(base^ret, rounding_digits)
    } else {
      by <- floor((max - min) / n) + 1
      ret <- base ^ seq(min, max, by = by)
    }
    ret
  }
}
```

This function yields the following results (compared to the current version in the scales package).

```r
> log_breaks(base=exp(1))(c(2.0001, 2.000111111))
[1] 2.000100 2.000103 2.000106 2.000108 2.000111
> scales::log_breaks(base=exp(1))(c(2.0001, 2.000111111))
[...]

PR #117, opened 2018-06-01
log_breaks() add finer breaks when needed

ThierryO:
The change is only effective when few brakes are with the limits. The effect is best shown by the examples below. Visible breaks (= shown on the plot) are in bold. Note that the original version displays only one break or even no breaks at all. The intermediate values are chosen so that the cut the largest gap between the brakes is two somewhat equally sized chunks. E.g. $log_10(3) = 0.477$ Hence 3 is used to split `c(1, 10)`. 

`log_breaks()(c(1664, 14008))`
- original breaks: 1,000 **10,000** 100,000
- new breaks: 1,000 **3,000** **5,000** **10,000** 30,000 50,000 100,000 300,000 500,000

`log_breaks()(c(407, 3430))`
- original breaks 100   **1,000**  10,000
- new breaks 100   300   **500**  **1,000**  **3,000**  5,000 10,000 30,000 50,000

`log_breaks()(c(1761, 8557))`
- original breaks: 1,000  10,000
- new breaks: 1,000  **2,000**  **3,000** **5,000** 10,000 20,000 30,000 50,000


hadley:
Fixes #100 

@ThierryO do you have thoughts on how to test this?

@dpseidel can you please review this approach?

ThierryO:
I've added some unit tests and merged the changes from the master branch

dpseidel:
This is a nice solution. There is a small bug however for `base <= 2`: due to the way candidate intermediate values are set you can end up with duplicate breaks.  

```
# master log_breaks
scales::log_breaks(n = 5, base = 2)(c(3, 100))
#> [1]   2   8  32 128

# PR#117  log_breaks
log_breaks(n = 5, base = 2)(c(3, 100))
#> [1]   2   4   4   8   8  16  16  32  32  64  64 128 128 256

# happens on any other interval that demands more relevant breaks
log_breaks(base = 2, n = 10)(c(.2, 5))
#>  [1]  0.125  0.250  0.250  0.500  0.500  1.000  1.000  2.000  2.000  4.000
#> [11]  4.000  8.000  8.000 16.000
```

Can you fix this and add another a test for base2 to ensure it's working as expected?

R/breaks.r
```diff
@@ -59,7 +62,27 @@ log_breaks <- function(n = 5, base = 10) {
...
+    breaks <- base ^ seq(min, max, by = by)
+    relevant_breaks <- base ^ rng[1] <= breaks & breaks <= base ^ rng[2]
+    if (sum(relevant_breaks) >= (n - 2)) return(breaks)
+
+    steps <- 1
+    delta <- function(x) {
+      min(diff(log(sort(c(x, steps, base)), base = base)))
+    }
+    candidate <- 2:(base - 1)
+    while (length(candidate)) {
```
hadley: I wonder if we should extract this out into a named function?

R/breaks.r
```diff
@@ -59,7 +62,27 @@ log_breaks <- function(n = 5, base = 10) {
...
+      best <- which.max(sapply(candidate, delta))
```
hadley: Can you please use `vapply()` here?

R/breaks.r
```diff
@@ -59,7 +62,27 @@ log_breaks <- function(n = 5, base = 10) {
...
+      steps <- c(steps, candidate[best])
+      candidate <- candidate[-best]
+
+      breaks <- as.vector(outer(base ^ seq(min, max), steps))
+      relevant_breaks <- base ^ rng[1] <= breaks & breaks <= base ^ rng[2]
+      if (sum(relevant_breaks) >= (n - 2)) {
+        break
+      }
+    }
+    return(sort(breaks))
```
hadley: No need for explicit return here

R/breaks.r
hadley: Also would you mind adding a comment that explains the basic idea of the approach?

ThierryO:
@dpseidel : I've added a unit test for the base 2 problem and solved it

@hadley : I' updated the coding style and added some comments. Just a few question before I turn my addition into a separate function:
- should it be exported or not?
- should it return a function like `log_breaks()` does, or rather the `breaks`
- is `log_sub_breaks()` a good name?

ThierryO:
I've noticed that a bunch of unrelated Rd files have changed when running the checks. The checks yield several warnings, all related to `prepare_Rd: colour_ramp.Rd:38: unexpected macro '\link', expecting '{'`

hadley:
It should not be exported; it's just a helper to make the code a little clearer to read. I'm not sure about the other details, it'll probably need a little experimentation to get a sense for what the most natural decomposition is.

R/breaks.r
```diff
@@ -59,8 +62,70 @@ log_breaks <- function(n = 5, base = 10) {
...
+      by <- by - 1
+      breaks <- base ^ seq(min, max, by = by)
+      relevant_breaks <- base ^ rng[1] <= breaks & breaks <= base ^ rng[2]
+      if (sum(relevant_breaks) >= (n - 2)) return(breaks)
+    }
+    log_sub_breaks(rng, n = n, base = base)
+  }
+}
+
+#' Calculate intermediate log-scale breaks
```
hadley: Needs `@noMd` since we're not exporting.

hadley:

@dpseidel can you please take a pass through for style/readability; feel free to merge when you think it's ready.

R/breaks.r
ThierryO: The `@md` / `@noMd` tags turn Markdown parsing on / off for the given block. See https://cran.r-project.org/web/packages/roxygen2/news.html.

Is there another tag to remove the help file? Or maybe render is to a `non-exported` help file?

hadley: Uh, typo, sorry - I meant `@noRd`

ThierryO: done in e043d03

ThierryO:
@dpseidel you can finish this. Feel free to rewrite the description if needed. I'm not a native English speaker.

dpseidel:
While tidying the description, I noticed that the `log_sub_breaks` function breaks when fed `base = 2`. In fact, for all values `base < 3`, `length(candidate) == 0`, the `while` loop never runs  and the user is returned an error saying `Error in sort(breaks) : object 'breaks' not found ` ([lines 118:120](https://github.com/r-lib/scales/pull/117/files#diff-29253b9a84104413ea078e62c427870eR118-120)). 

Example:
```r
log_breaks(base = 2, n = 10)(c(101, 9999))
#>  Error in sort(breaks) : object 'breaks' not found 
```

Since intermediate breaks aren't often required for base 2, it can be hard to catch. Unless the user gives `log_breaks` a vector with a trivially small range or requests a large number of breaks relative to the range of the data, `log_sub_breaks` won't be called. 

Although a rare edge case, I imagine we want the `log_sub_breaks` function to be able to handle `base = 2` one way or another. The basic hack I made while testing was simply to set `breaks` before the `while` loop to the default `base ^ seq(min, max, steps)` where `steps = 1` , but this obviously negates the entire algorithm. As any solution offering intermediate steps to bases `< 3` will require non-integer steps, perhaps it's more in keeping with the philosophy behind the current approach to limit the use of `log_sub_breaks` to only bases `>=3`. Thoughts?

dpseidel:
An unrelated observation, this algorithm (due to the `relevant breaks` condition) often returns a lot more breaks beyond the range of the given data. Given that `log_breaks()` is most often called as a part of setting the axis in `ggplot2`, this behavior is probably of no importance --- breaks beyond the range of the data have no affect on the plots --- but I thought I would note the difference from current behavior. It could be pretty easily prevented by returning just the relevant breaks or relevant breaks plus one on either side.

ThierryO:
@dpseidel the `base <= 2` problem was reintroduced with 28f404e and is now fixed in 
e825ec5. In case of `base <= 2` there are no integer values between 1 and `base`, hence the function returns only the integer powers of `base`.

The breaks (way) beyond the range of the data was due to the combination of me knowing that ggplot2 ignores them and me being lazy. This is now fixed in 6f0f442. The function returns no more that one break outside of the range on each side of the range.

dpseidel:
Thanks @ThierryO! I appreciate all your work on this.

hadley:
Looks great. I really appreciate all your work on this @ThierryO — it's going to improve a lot of plots made with ggplot2!

ThierryO:
You're welcome. I am glad that part of my code is also useful for others.

sjackman:
Thanks for this work, Thierry!

I think there's a minor improvement possible when it selects three sub breaks. If I understand the description, it'll select 1-3-5, where as the [1-2-5 series](https://en.wikipedia.org/wiki/Preferred_number#1–2–5_series) of convenient/preferred has more even spacing. The gap between 1 and 3 is 80% larger than the size of the gap from 3 to 5. Whereas the gap between 2 to 5 is only 25% larger than the gap from 1 to 2.

3/1 = 3; 5/3 = 1.67; 10/5 = 2
whereas
2/1 = 2; 5/2 = 2.5 ; 10/5 = 2

Demonstrated by these plots

# 1-2-5 series

image

# 1-3-5 series

image

```r
library(scales)
library(tidyverse)

log_breaks_125 <- function(...) c(1, 2, 5) %o% log_breaks()(...)
log_breaks_135 <- function(...) c(1, 3, 5) %o% log_breaks()(...)

ggplot(diamonds, aes(x = carat, y = price)) +
	scale_x_log10(breaks = log_breaks_125) +
	scale_y_log10(breaks = log_breaks_125) +
	geom_point()

ggplot(diamonds, aes(x = carat, y = price)) +
	scale_x_log10(breaks = log_breaks_135) +
	scale_y_log10(breaks = log_breaks_135) +
	geom_point()
```

Thanks again for this work. It's certainly much better than before, with no sub breaks.

ThierryO:
The algorithm does not search for the length `n` series which is the best approximation of equidistant breaks.

The algorithm works sequential. It first looks at the best integer to split the 1 - `base` interval. That is 3 when `base = 10`. Then it looks for the next integer to add. Adding another integer will spilt either the 1 - 3 interval or the 3 - 10 interval. 3 - 10 is the largest interval and 5 is the best split for that interval. The benefit of this algorithm is that is works for any `base > 2`, for up to `base - 1` number of breaks and it is straightforward to add extra integers.

Having it to return the 1 - 2 - 5 series will require a complete overhaul of the algorithm.

sjackman:
Here's a possibility that aims to find equidistant breaks.

```r
log_subbreaks <- function(n = 3, base = 10) unique(round(base ^ (0:(n-1)/n)))
lapply(1:5, log_subbreaks)
```
[[1]]
[1] 1

[[2]]
[1] 1 3

[[3]]
[1] 1 2 5

[[4]]
[1] 1 2 3 6

[[5]]
[1] 1 2 3 4 6
```

The above code isn't quite right, because `log_subbreaks(n = 10, base = 10)` returns only seven breaks `1 2 3 4 5 6 8`, but you get the idea.

PR #124, opened 2018-06-06
Minor documentation edits

dpseidel:
Mostly small edits to links to resolve warnings when rebuilding documentation.
Various other tidying edits the result of #121.

### DIFF
--- a/R/colour-mapping.r
+++ b/R/colour-mapping.r
@@ -350,7 +350,7 @@ filterRange <- function(f) {
 
 #' Fast color interpolation
 #'
-#' Returns a function that maps the interval [0,1] to a set of colors.
+#' Returns a function that maps the interval \[0,1] to a set of colors.
 #' Interpolation is performed in the CIELAB color space. Similar to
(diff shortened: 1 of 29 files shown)


R/colour-mapping.r
```diff
@@ -373,7 +373,7 @@ filterRange <- function(f) {
 #' @return A function that takes a numeric vector and returns a character vector
 #'   of the same length with RGB or RGBA hex colors.
 #'
-#' @seealso \link[grDevices]{colorRamp}
+#' @seealso \code{\link[grDevices]{colorRamp}}
```
hadley: I think this should be `[grDevices:colorRamp]`

R/pal-dichromat.r
```diff
@@ -1,5 +1,6 @@
 #' Dichromat (colour-blind) palette (discrete).
 #'
+#' @noMd
```
hadley: What do you need this for?

dpseidel: without it the \Sexpr that follows breaks - seems to be the same bug documented 
 in klutometis/roxygen#682.

hadley: Bummer

PR #135, opened 2018-06-27
Fix boxcox_trans equations and docs

### LINKED ISSUE #103: boxcox_trans functions not always valid
In boxcox_trans, trans and inv functions include sign() and abs() (maybe for preserving order of values when 0<x<1 ?), but this then passes not totally correct functions to the transformer. 
```
 trans <- function(x) (x^p - 1)/p * sign(x - 1)
inv <- function(x) (abs(x) * p + 1 * sign(x))^(1/p)
```
Using this form, for data 0 < x < 1,  inv(trans(x)) != x

Eliminating those and using the basic formula for Box-Cox Power transformations:
```
trans <- function(x) (x^p - 1)/p
inv <- function(x) (x * p + 1)^(1/p)
```
I've confirmed this works for all of my data x>0, and that inv(trans(x)) == x. I've also used this to scale my axes successfully.
I'm new to git, so I'm just reporting this here.

dpseidel:
Fixes #103 

Our previous equations were not invertible for values less than or equal to 1.  I have adapted the equations to match published documentation of the Box-Cox power transformation and added tests. 

Additionally I fixed an error in the documentation and added details emphasizing that the BoxCox power transformation is intended for strictly positive values.


hadley:
Didn't we discuss using the modulus transform as a generalisation of Box-Cox that does work for negative values?

dpseidel:
I don't remember discussing it but I can look into it.

hadley:
Paper at https://www.jstor.org/stable/2986305

hadley:
Implementation at https://github.com/hadley/bigvis/blob/master/R/mt.r

hadley:

(Then feel free to merge)

PR #137, opened 2018-06-28
Update documentation for date_format

dpseidel:
Export `time_format` and combine documentation with `date_format`. A followup to #136 . 

Currently code for time_format and date_format are colocated in `trans-date.r` file as that is where `date_format` was original written. Given the addition of time_format, do we want to move them to the `formatter.r` file with other formatters or into a standalone file?


hadley:
If there's already a formatters file, lets put them there!

R/formatter.r
```diff
@@ -314,3 +314,34 @@ unit_format <- function(unit = "m", scale = 1, sep = " ", ...) {
...
   }
 }
+
+#' Formatted dates and times.
+#'
+#' @param format Date/time format using standard POSIX specification.  See
+#'  [strptime()] for possible formats.
+#' @param tz a time zone name, see [timezones()]. Defaults
+#'  to UTC
+#' @export
```
hadley: Can has examples?

dpseidel: 👍 done. Also removed the stop_if from tests because I realized it was unnecessary, as a suggests it is loaded.

PR #138, opened 2018-07-02
Provide separate box-cox and modulus trans functions

dpseidel:
As suggested by @statist7 in #103, exporting modulus and Box-Cox transformations as separate functions. `boxcox_trans` now gives an informative warning when given negative values.


NEWS.md
```diff
@@ -1,7 +1,9 @@
 # scales 0.5.0.9000
 
-* `boxcox_trans` now implements the modulus transformation, a generalisation
-  appropriate for data with both positive and negative values (@dpseidel, #103).
+* New function `modulus_trans` implements the modulus transformation for positive
+  and negative values (@dpseidel).
+
+* `boxcox_trans` is now invertible for `x >= 0` (@dpseidel, #103).
```
hadley: Missing parens after function name

Mention only works for positive values.

R/trans-numeric.r
```diff
@@ -16,17 +16,17 @@ atanh_trans <- function() {
   trans_new("atanh", "atanh", "tanh")
 }
 
-#' Box-Cox (modulus) power transformation.
+#' Modulus transformation.
```
hadley: Document modulus and boxcox together?

Maybe include the equations in the description?

R/trans-numeric.r
```diff
@@ -46,6 +46,36 @@ boxcox_trans <- function(p) {
...
+#' @references Box, G. E., & Cox, D. R. (1964). An analysis of transformations.
+#' Journal of the Royal Statistical Society. Series B (Methodological), 211-252.
+#' \url{https://www.jstor.org/stable/2984418}
+#'
+#' @export
+boxcox_trans <- function(p) {
+  if (abs(p) < 1e-07) return(log_trans())
+
+  trans <- function(x) {
+    if (any(x < 0)) {
```
hadley: Perfect, thanks!

statist7: Can `boxcox_trans` include `offset = 0`as an argument, to allow both versions of the Box-Cox transformation, see #151 ?

The modulus transformation as described by John and Draper assumed `offset = 1`, but in reality it ought to be general. So I suggest adding it as an argument to `modulus_trans` too.

R/trans-numeric.r
```diff
@@ -16,17 +16,56 @@ atanh_trans <- function() {
   trans_new("atanh", "atanh", "tanh")
 }
 
-#' Box-Cox (modulus) power transformation.
```
hadley: and modulus

R/trans-numeric.r
```diff
@@ -16,17 +16,56 @@ atanh_trans <- function() {
...
+
+  inv <- function(x) (x * p + 1)^(1 / p)
+
+  trans_new(
+    paste0("pow-", format(p)), trans, inv
+  )
+}
+
+#' @rdname boxcox_trans
+#' @details The modulus transformation implements a generalisation of the
```
hadley: I think it's better to put all the details in the first doc block. That makes it easier to check that it reads well

R/trans-numeric.r
hadley: That feels slightly too overparameterised to me, given how few people know about the modulus transformation.

statist7: I agree that few people know of the modulus transformation, including me until a few weeks ago when I saw your mention of it.

But that said, if we are pointing people with negative data towards it, is it not incumbent on us to highlight how it differs from `boxcox?` As well as the sign difference there is also the offset difference, which is far from obvious unless pointed out, and adding an `offset` parameter would do that. Its addition would also make both `boxcox` and `mt` much more flexible as transformations.

If I can't sell you the idea of the offset parameter for `mt`, do please consider it for `boxcox` as it is an important element of Box & Cox (1964).

hadley: Ok, that sounds reasonable. @dpseidel can you please add to both?

dpseidel:
Alrighty - offset parameter has been implemented!

R/trans-numeric.r
```diff
@@ -16,36 +16,79 @@ atanh_trans <- function() {
...
-#' @references J. John and N. Draper. "An alternative family of
-#' transformations." Applied Statistics, pages 190-197, 1980.
+#' Transformations used to transform non-normal data
+#' to more closely approximate a normal distribution.
+#'
+#' @param p Transformation exponent, \eqn{\lambda}.
+#' @param offset Constant offset. 0 for Box-Cox type 1,
+#'   otherwise any non-negative constant (Box-Cox type 2). `modulus_trans()`
+#'   sets the default to 1.
+#' @details The Box-Cox power transformation (type 1) requires strictly positive values and
```
hadley: Can you move this under the first paragraph? (and hence remove `@details`

R/trans-numeric.r
```diff
@@ -16,36 +16,79 @@ atanh_trans <- function() {
...
 
-#' Box-Cox (modulus) power transformation.
+#' Box-Cox & modulus transformations.
 #'
-#' @param p Exponent of Box-Cox transformation.
-#' @details Implements a generalisation of the Box-Cox transformation that works
-#' for data with both positive and negative values.
-#' @references J. John and N. Draper. "An alternative family of
-#' transformations." Applied Statistics, pages 190-197, 1980.
+#' Transformations used to transform non-normal data
```
hadley: I'd say something like: The Box-Cox transformation is a flexible transformation, often used to transform data towards normality. The modulus transformation generalises Box-Cox to also work with negative values.

statist7:
Thank you both!

Issue #140, opened 2018-07-04
scales::percent adds comma's


markvanderloo:
Saw this pop up via twitter.

```r
> scales::percent(1e4)
[1] "1,000,000%"
```
two issues:

1. Comma separator is not documented.
2. The comma as 'big' separator is biased to the North-American convention. In European countries (including NL), we use comma as decimal separator and dot as big separator (especially for, but not limited to financial amounts). 

Would be nice to have an option like `big.mark` behaving like the same option in `base::format`.

hadley:
Related to #79, and should be part of #77

Issue #128, opened 2018-06-13
percent: normal values get rounded to 0% or 100% when mixed with huge values


CesarSancho:
I am using percent from scales 0.5.0. If I apply it to a vector containing big numbers, the "normal" ones get rounded to zero, see this example:

> percent(c(0.8, 1e2))
[1] "80%"     "10,000%"
> percent(c(0.8, 1e3))
[1] "100%"     "100,000%"
> percent(c(0.8, 1e4))
[1] "0%"         "1,000,000%"

I am using r 3.4.2 on windows:
> sessionInfo()
R version 3.4.2 (2017-09-28)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1

dpseidel:
This is addressed in PR #77. Once it's merged, the `percent` and `percent_format` functions will have a new `accuracy` argument to help users control this rounding behavior. So stay tuned, we hope to get this merged soon!

CesarSancho:
thank you Dana

I have made a patch for myself; I am not sure if it is worth contibuting and I would not know how to do it, so I paste it here below in case somebody find it useful. I just make sure that the minimum precision will be at least 1%, so numbers are not unnecesarily rounded. This fixes my issue without having to add parameters in the "percent" call:

```r
percent <- function (x){
  if (length(x) == 0) 
    return(character())
  #  x <- round_any(x, myPrecision(x)/100)
  x <- round_any(x, min( myPrecision(x)/100, 0.01) )
  paste0(comma(x * 100), "%")
}
```

Issue #76, opened 2016-03-03
should percent_format allow variable precision?


AndrewsOR:
I've noticed that when using `percent(x)`, if elements of x are of of different orders of magnitude, percent will pick the precision that works for the largest element.  

```
> percent(c(0.123,0.000234))
[1] "12.3%" "0.0%" 
```

In other words, precision is constant -- the number of significant figures is not.  But sometimes when making a decision tree or a table for a report it's nice to have constant significant figures.

The [workaround ](http://stackoverflow.com/questions/34934374/percentage-with-significant-digits) I currently use is easy enough:

```
> sapply(c(0.123,0.000234), percent)
[1] "12.3%"  "0.0234%"
```

to ensure reverse compatibility, maybe something like

```
percent(x, adjust=FALSE)
```

where `adjust=TRUE` would ensure a consistent number of significant figures

danwwilson:
Any chance of merging #77 into the master or addressing it by extending the code? It would be great to be able to change the `big,mark` and `decimal.mark` at the very least to assist with European representation of numbers.

Issue #72, opened 2016-02-22
Specify the number of rounding places for percent


derek-damron:
I have long been a fan of the `percent` function but there have been instances where I have wanted some more control over the number of rounding places of the output.

This could take a few different forms, so below is a first pass to illustrate the idea.  There is almost certainly a better way to formalize the `nplaces` argument to make it more consistent with the rest of the **scales** package.

``` r
#' @param nplaces The number of decimal places to round the percent.  
#'   For example, nplaces=2 will convert 0.25121 into "25.12%".
percent_format <- function(nplaces = NULL) {
  function(x) {
    if (length(x) == 0) return(character())
    if (is.null(nplaces)) x <- round_any(x, precision(x) / 100)
    else x <- round_any(x, 10^-(nplaces+2))
    paste0(comma(x * 100), "%")
  }
}
```

Some examples of this change in action:

```
> percent_format()(.01234567)
[1] "1.23%"
> percent_format(1)(.01234567)
[1] "1.2%"
> percent_format(3)(.01234567)
[1] "1.235%"
```

What do you think about this addition?

PR #146, opened 2018-07-06
updating comma_format, percent_format and unit_format

larmarange:
now based on number_format()

cf. discussion in https://github.com/r-lib/scales/pull/142


larmarange:
Solve:

- https://github.com/r-lib/scales/issues/140
- https://github.com/r-lib/scales/issues/128
- https://github.com/r-lib/scales/issues/76
- https://github.com/r-lib/scales/issues/72

larmarange:
Notes:

- `NEWS.md` not yet updated
- for `unit_format` and `percent_format`, I kept the default value of `big.mark` used in `number_format` (i.e. a space) for consistency
- for `unit_format`, I kept `unit` and `sep` arguments for backward compatibility. Maybe it could be relevant to add in documentation that `unit_format` is deprecated and that `number_format` could be used instead.

hadley:
Can you please use `Fixes #140`? That specific formulation will cause the related issue to be closed when this PR is merged.

R/formatter.r
```diff
@@ -1,6 +1,8 @@
-#' Number formatter: a generic formatter for numbers
+#' Number formatters
 #'
-#' @return `number_format` returns a function with single parameter
+#' `number_format` and `number` are generic formatters for numbers.
```
hadley: Can you please use `()` for functions?

R/formatter.r
```diff
@@ -74,32 +72,25 @@ number <- function(x, accuracy = 1, scale = 1, prefix = "",
   ret
 }
 
-#' Comma formatter: format number with commas separating thousands.
-#'
-#' @param ... other arguments passed on to [format()]
-#' @param x a numeric vector to format
-#' @return a function with single parameter x, a numeric vector, that
-#'   returns a character vector
+#' `comma_format` and `comma` format numbers with commas separating thousands.
```
hadley: I think it's easier to understand if you put all the description in one place. 

(But the other tags should stay with individual functions)

R/formatter.r
```diff
@@ -74,32 +72,25 @@ number <- function(x, accuracy = 1, scale = 1, prefix = "",
...
 #' comma(c(1, 1e3, 2000, 1e6))
 #'
-#' # If you're European you can switch . and , with the more general
-#' # format_format
-#' point <- format_format(big.mark = ".", decimal.mark = ",", scientific = FALSE)
-#' point(c(1, 1e3, 2000, 1e6))
-#' point(c(1, 1.021, 1000.01))
-comma_format <- function(...) {
+comma_format <- function(big.mark = ",", ...) {
   force_all(...)
```
hadley: I think this force all can be removed.

R/formatter.r
```diff
@@ -182,25 +173,50 @@ dollar <- function(x, prefix = "$", suffix = "",
...
+#' per_mille(.1234)
+#'
+#' french_percent <- percent_format(
+#'   decimal.mark = ",",
+#'   suffix = " %"
+#' )
+#' french_percent(runif(10))
+#'
+percent_format <- function(accuracy = NULL, scale = 100, suffix = "%", ...) {
+  force_all(...)
```
hadley: `force_all()` is only needed in functions that return functions.

R/formatter.r
```diff
@@ -182,25 +173,50 @@ dollar <- function(x, prefix = "$", suffix = "",
...
+#' )
+#' per_mille(.1234)
+#'
+#' french_percent <- percent_format(
+#'   decimal.mark = ",",
+#'   suffix = " %"
+#' )
+#' french_percent(runif(10))
+#'
+percent_format <- function(accuracy = NULL, scale = 100, suffix = "%", ...) {
```
hadley: Is there a reason to make scale or suffix parameters?

R/formatter.r
```diff
@@ -74,32 +72,25 @@ number <- function(x, accuracy = 1, scale = 1, prefix = "",
...
-#' @return a function with single parameter x, a numeric vector, that
-#'   returns a character vector
+#' `comma_format` and `comma` format numbers with commas separating thousands.
 #' @export
+#' @rdname number_format
 #' @examples
+#' # comma() and comma_format()
 #' comma_format()(c(1, 1e3, 2000, 1e6))
-#' comma_format(digits = 9)(c(1, 1e3, 2000, 1e6))
+#' comma_format(accuracy = .01)(c(1, 1e3, 2000, 1e6))
```
hadley: We probably do need to preserve the digits argument for backward compatibility.

R/formatter.r
hadley: Oh, I see for `per_mille`. I think it would be simpler to continue to use `number_format()` for that example.

R/formatter.r
```diff
@@ -372,27 +388,28 @@ precision <- function(x) {
...
 #' ha <- unit_format(unit = "ha", scale = 1e-4)
 #' km(runif(10) * 1e5)
-#' @seealso [comma()]
-unit_format <- function(unit = "m", scale = 1, sep = " ", ...) {
-  force_all(unit, scale, sep, ...)
-  function(x) {
-    paste(comma(x * scale, ...), unit, sep = sep)
-  }
+#'
+unit_format <- function(unit = "m", sep = " ", suffix = paste0(sep, unit), ...) {
```
hadley: I think it's still worth including scale here, just to be explicit.

R/formatter.r

R/formatter.r
larmarange: ok. I have also reordered the code (i.e. each function being declared in the R file in the same order as the doc).

R/formatter.r
larmarange: ok it has been removed.

by security, I added it in `number_format()`

R/formatter.r

larmarange:
**Regarding allowing `...` in `percent format()`**

If I have to use `number_format` for displaying percentage with a comma as a decimal separator, I need to re-declare all options, i.e. `number_format(decimal.mark = ",", scale = 100, accuracy = NULL, suffix = "%")` while using `percent(big.mark = ",")` is much easier and intuitive.

**Regarding digits argument in `comma_format()`**

In the current implementation of `comma_format()`, `digits` should be understood as in `format()` (i.e. total significant numbers to display, before or after the decimal point) and not the number of figures after the decimal point as in `round()`.

```r
> comma(123.45678)
[1] "123.4568"
> comma(123.45678, digits = 0)
[1] "123"
> comma(123.45678, digits = 2)
[1] "123"
> comma(123.45678, digits = 5)
[1] "123.46"
```

That behaviour is not very intuitive. 

In addition, it will require some code to determine the corresponding value of `accuracy` to be passed on `number_format` or it will require some changes in `number_format()` by adding a `digits` argument and to not take into account the `accuracy` argument if `digits` is supplied. Not sure that there is a lot of situations where `digits` in that sense.

I could understand a `digits` argument corresponding to the number of figures to display after the decimal point (like in `round()`) and then to fix `accuracy` accordingly, but such behaviour will not maintain backward compatibility for `comma`.

**Regarding the `...` argument for other formatters than `number_format`**

If we keep in mind that `ordinal_format` or `dollar_format` could also rely on `number_format` with some extra functionalities, then it will be relevant to keep `...` for those functions. Wouldn't it be more consistent to keep `...` in all formatters then?

hadley:
* Ok.

* Ok. In that case, we should formally deprecate the argument, `warning()` if it is `!missing()`, and suggesting to use `accuracy` instead

* The alternative is to explicitly pass along every argument. This is more work for the developer, but it makes life for easier for the user (because they get autocomplete on function arguments, and the documentation doesn't need to be read as closely to figure out what happens with `...`)

larmarange:
I have updated all functions to pass all arguments.

A warning has been added to `comma()` and `comma_format`.

R/formatter.r
```diff
@@ -74,32 +76,151 @@ number <- function(x, accuracy = 1, scale = 1, prefix = "",
...
-#' point(c(1, 1e3, 2000, 1e6))
-#' point(c(1, 1.021, 1000.01))
-comma_format <- function(...) {
-  force_all(...)
-  function(x) comma(x, ...)
+comma_format <- function(accuracy = 1, scale = 1, prefix = "",
+                         suffix = "", big.mark = ",", decimal.mark = ".",
+                         trim = TRUE, digits, ...) {
+  if (!missing(digits)) {
+    warning("'digits' argument is deprecated, use 'accuracy' instead.")
```
hadley: Needs `call. = FALSE`, and instead of `'digits'` can you please do `` `digits` ``? (And similarly for accuracy)

R/formatter.r
```diff
@@ -1,6 +1,12 @@
-#' Number formatter: a generic formatter for numbers
+#' Number formatters
 #'
-#' @return `number_format` returns a function with single parameter
+#' `number_format()` and `number()` are generic formatters for numbers.
```
hadley: Maybe add something like: ``"All formatters allow you to re-`scale` (multiplicatively) and round to specified `accuracy`"``

R/formatter.r

larmarange:
All fixed and news bullet added

R/formatter.r
```diff
@@ -1,6 +1,16 @@
-#' Number formatter: a generic formatter for numbers
+#' Number formatters
 #'
-#' @return `number_format` returns a function with single parameter
+#' `number_format()` and `number()` are generic formatters for numbers.\cr
```
hadley: Please eliminate the `\cr` - these should all run into together in one paragraph. (Or if you don't like that, use a bulleted list instead)

hadley: Or maybe you want to use `@description` so you can use regular new lines?

R/formatter.r
```diff
@@ -16,26 +26,22 @@
...
-#'   scale = 1000,
-#'   suffix = "\u2030",
-#'   accuracy = .1
-#' )
-#' per_mille(v)
 number_format <- function(accuracy = 1, scale = 1, prefix = "",
-                          suffix = "", big.mark = " " , decimal.mark = ".",
+                          suffix = "", big.mark = " ", decimal.mark = ".",
                           trim = TRUE, ...) {
+  force_all(...)
```
hadley: Needs to include all args

R/formatter.r

R/formatter.r

larmarange:
merge conflict solve and corrections made

hadley:
Thanks!

PR #149, opened 2018-07-07
New arguments for `ordinal_format`

larmarange:
`prefix`, `suffix`, `big.mark` and `rules` (allowing to provide custom rules)


R/formatter.r
```diff
@@ -232,32 +232,70 @@ scientific <- function(x, digits = 3, ...) {
...
 #' @export
 #' @examples
 #' ordinal_format()(1:10)
 #' ordinal(1:10)
-ordinal_format <- function() {
-  function(x) ordinal(x)
+#'
+#'  # Custom rules for French
+#' french <- list(
+#'   er = "^1$",
```
hadley: Does this work for 101? In English it would be 101st, but in French 101e?

R/formatter.r
```diff
@@ -232,32 +232,70 @@ scientific <- function(x, digits = 3, ...) {
...
+#'   nd = "^2$",
+#'   e = "([0-9]+)[12]$",
```
hadley: What happens if a number matches multiple regular expressions? If it return the first match, we could make the fallbacks (in both French and English) simpler by matching anything (i.e. `e = '.'`)

R/formatter.r
```diff
@@ -232,32 +232,70 @@ scientific <- function(x, digits = 3, ...) {
...
 #' Ordinal formatter: add ordinal suffixes (-st, -nd, -rd, -th) to numbers.
 #'
-#' @return a function with single parameter x, a numeric vector, that
+#' @return A function with single parameter `x`, a numeric vector, that
 #'   returns a character vector
-#' @param x a numeric vector to format
+#' @param x A numeric vector of positive values to format.
+#' @param prefix,suffix Symbols to display before and after value.
+#' @param big.mark Character used between every 3 digits to separate thousands.
+#' @param rules Custom rules for computing ordinal indicators (list of regular
```
hadley: Maybe: Named list of regular expressions, match in order. Name gives suffix, and value specifies which numbers to match.

R/formatter.r
```diff
@@ -232,32 +232,70 @@ scientific <- function(x, digits = 3, ...) {
...
 }
 
 #' @export
 #' @rdname ordinal_format
-ordinal <- function(x) {
+ordinal <- function(x, prefix = "", suffix = "", big.mark = " ",
+                    rules = NULL, ...) {
   stopifnot(all(x > 0))
+  if (is.null(rules)) {
+    rules <- list(
```
hadley: Maybe we should make this `ordinal_english()` and export it? Then we can use as the default argument, and users can easily see what the default rules are.

R/formatter.r
larmarange: Yes, in French it is "101e" and "1001e", "er" is used only for 1.

R/formatter.r
larmarange: I have done a quick test. If a number match two or more rules, two or more values will be returned. I will have to update the code to consider only the first results

larmarange: corrected

R/formatter.r

R/formatter.r

also added `ordinal_french()` and `ordinal_spanish()`

tests/testthat/test-formatter.r
```diff
@@ -82,6 +82,18 @@ test_that("ordinal format", {
   expect_equal(ordinal(21), "21st")
```
hadley: Can you please add a test for 101st, too?

R/formatter.r
```diff
@@ -232,32 +232,98 @@ scientific <- function(x, digits = 3, ...) {
...
   )
+}
+
+#' @export
+#' @rdname ordinal_format
+ordinal_french <- function() {
+  list(
+    er = "^1$",
+    nd = "^2$",
+    e = "([0-9]+)[12]$",
```
hadley: Now I think you can simplify this so that `e` matches anything

hadley: And similarly, English just needs a fall back to `th` if nothing else matches.

larmarange:
done, bullet news added and merge conflict solved

PR #69, opened 2016-01-12
"byte" formatter

hrbrmstr:
Five new functions, 
- `byte_format` (convert numeric vector to byte measurement labels such as 1024 => "1 Kb")
- `bytes` (the one that does all the work)
- 3 "shorthand" functions: `Kb`, `Mb` & `Gb`

These support binary (1024-base) and SI (1000-base) conversions.

This will enable users to avoid rolling their own or adding `gdata` as a dependency and make "cyber" a first-class citizen since pretty much everyone uses `scales` :-) 

If it'd be better without the three shorthand functions I can remove them and do a new PR.

PR #156, opened 2018-07-23
Byteformatter

dpseidel:
Implementing the byte formatter dreamed up by @hrbrmstr (#69), now with compatibility with the generic number formatter implemented recently. 

------
Also, perhaps a separate issue but I noticed in working on this that the default `big.mark` for `number()` and `percent()` is a space, giving kind of an awkward output: 
``` r
library(scales)
number(c(1000, 2000, 4000))
#> [1] "1 000" "2 000" "4 000"
percent(c(2, 5, 20))
#> [1] "200%"   "500%"   "2 000%"
```
I'd like to change that to default `big.mark = ""` -- agreed? Can I slide that in here, or should I do that in a separate PR?


hadley:
Lets discuss `big.mark` separately.

hadley:
And make sure to add a `closes #69` somewhere

PR #161, opened 2018-07-24
Adjust expand_range() so zero range scales respect mul and add

dpseidel:
This seemed the most logical and minimal fix but if you have a different idea, let me know.  Previously in expand_range(), `zero_width` was divided in half and added or subtracted on either side of the axis tick, resulting (from defaults) in a range of 1 (so when the original issue poster called `geom_boxplot(width=0)` there was no padding). Now instead, `zero_width` functions as our `diff(range)` and the equation matches that of non-zero ranges, respecting `mul` and `add`. 

This fixes (the scales side of) the original issue (respecting `mul` and `add` parameters) but single ticket facets specifying a boxplot `width` ≥  than this range will still look tight. Looking into this further I found some weird bugs happening with the `width` argument to `geom_boxplot` and `geom_violin`. Will open a separate issue wherever appropriate once I have a better grasp on what's going on. 

Fixes tidyverse/ggplot2#2281. 

Does this deserve a news bullet for scales? What sorts of downstream tests should I run? Would those belong on this PR?

### DIFF
--- a/R/bounds.r
+++ b/R/bounds.r
@@ -228,7 +228,7 @@ expand_range <- function(range, mul = 0, add = 0, zero_width = 1) {
   if (is.null(range)) return()
 
   if (zero_range(range)) {
-    c(range[1] - zero_width / 2, range[1] + zero_width / 2)
+    range + c(-1, 1) * (zero_width * mul + add)
   } else {
     range + c(-1, 1) * (diff(range) * mul + add)
   }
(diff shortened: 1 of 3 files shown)


R/bounds.r
```diff
@@ -228,7 +228,7 @@ expand_range <- function(range, mul = 0, add = 0, zero_width = 1) {
   if (is.null(range)) return()
 
   if (zero_range(range)) {
```
hadley: Maybe extract out width to be something like:

```R
width <- if (zero_range(range)) zero_width else diff(range)
```

hadley:
Yes, needs a news bullet (can point to the ggplot2 issue)

I'd add a test just to make sure that mult and add affect `expand_range()` when the range is zero.

hadley:
(and please double check that it fixes the motivating problem)

PR #150, opened 2018-07-13
Build pkgdown site

dpseidel:
Closes #130 

A first pass. We may choose to wait to merge, or at a minimum, we will need to rebuild site after final PRs and issues are closed and before release. 

A couple notes:
- two independent fixes to pkgdown and roxygen made it possible to remove the `@noMd` tags
- I added the `Encoding` info to the DESCRIPTION file to silence the roxygen2 warning.

PR #163, opened 2018-07-25
Update website

dpseidel:
This should probably just sit and wait for a final rebuild after final PR merges and release, but this PR includes updates to reflect feedback from #150. 

I don't have admin access in r-lib to change config but the website should also be changed to  [scales.r-lib.org]()


hadley:
I set up http://scales.r-lib.org; it'll take a couple of minutes to go live.

PR #164, opened 2018-07-25
Add new si formatter

### LINKED ISSUE #83: add something like `nice_dollar`
I was writing this hack recently

```R
t_dollar <- dollar_format(suffix = 'T')
b_dollar <- dollar_format(suffix = 'b')
m_dollar <- dollar_format(suffix = 'm')
k_dollar <- dollar_format(suffix = 'k')

nice_dollar <- function(x) {
    sapply(x, FUN=function(amount) {
        if (is.na(amount)) return('n.a.')
        if (amount > 9e11) return(t_dollar(amount/1e12))
        if (amount > 9e8) return(b_dollar(amount/1e9))
        if (amount > 9e5) return(m_dollar(amount/1e6))
        if (amount > 9e2) return(k_dollar(amount/1e3))
        dollar(amount)
    })
}
```

which is useful in scenarios with free scales

image

dpseidel:
Closes #83

Adds a new function `nice_dollar()` that allows for variable `suffix` and `scale` based magnitude. 
```{r}
nice_dollar(c(1e3, 1e6, 1e9))
#> [1] "1K" "1M" "1B"
```

Might deserve a better name?

I think is a neat functionality but I went back and forth on whether, in this form, it was a flexible enough function or broad enough use case to include in the package. 

This would be the last change to code before release.


hadley:
I'd attack the problem differently:

```R
number_si <- function(x, prefix = "", suffix = "", ...) {
  
  breaks <- 10 ^ c(" " = 0, K = 3, M = 6, B = 9)
  
  n_suffix <- cut(x,
    breaks = c(-Inf, unname(breaks), Inf),
    labels = c(names(breaks)), 
    right = FALSE
  )
  
  scale <- 1 / breaks[suffix]
  
  number(x, scale = scale, suffix = paste0(suffix, n_suffix), ...)
}
```

It looks like `number()` is already vectorised so this will work.

(The code isn't quite right but it should get you in the right direction. I'd make sure to test negative inputs, inputs smaller than the first prefix, and larger than the largest)

(Also the name isn't quite right since these aren't quite SI prefixes)

hadley:
Or maybe we could somehow just pull it out into a pair of functions that you could use with the `scale` and `suffix` arguments of any formatter? i.e. `si_scale()`, `si_suffix()`

dpseidel:
So I think pulling out `si_scale()` or `si_suffix()` is a great idea, the idea being that it could allow variably scales and suffixes provided by the user right?  I would need to think a little more about how to code that up, but for now I have adapted the `number_si()` function and tested it for proper behavior with `Inf`, `NA`, negative, smaller, and larger values.  With some tidy eval we could even just adapt this `number_si()` function to do take custom inputs, right? But scales doesn't have rlang as a dependency so that probably isn't the right course. 

What do you think? Should we just finish up `number_si()` and get on with release? Or should I take a couple hours to sketch out the new functions? Or perhaps wait on this until after release?

hadley:
Lets leave this for now and focus on getting the release done.

PR #167, opened 2018-08-06
train_continuous now preserves the class of the input

### LINKED ISSUE #166: Enhance Supporting Nonstandard Classes in train_continuous
Could `train_continuous` also include an `if` block for the case when `existing` is `NULL` to better work with nonstandard classes??

``` r
train_continuous <- function(new, existing = NULL) {
  if (is.null(new)) return(existing)
  
  if (is.factor(new) || !typeof(new) %in% c("integer", "double")) {
    stop("Discrete value supplied to continuous scale", call. = FALSE)
  }
  if (is.null(existing)) {
    suppressWarnings(range(new, na.rm = TRUE, finite = TRUE))
  } else {
    suppressWarnings(range(existing, new, na.rm = TRUE, finite = TRUE))
  }
}
```

This issue relates to r-quantities/units#164.

When trying to use `train_continuous` with a nonstandard class, it ends up calling `range(existing, new)` (https://github.com/r-lib/scales/blob/3c17ebf8a9babd61b0e7b999aae83ee08e2f03f1/R/scale-continuous.r#L41).  With classes beyond base classes (in the case listed above, the 'units' class), this causes the class to be lost and yields issues down the line when using classes that cannot be directly compared.

billdenney:
Fix #166


NEWS.md
```diff
@@ -67,6 +67,9 @@
 
 * `rescale_mid()` now properly handles NAs (@foo-bar-baz-qux, #104).
 
+* `train_continuous()` now maintains the class of inputs when they are not numeric
```
dpseidel: @hadley is this going to make the release? I suspect it won't and this should actually be in its own new section up top: 
```
# scales 1.0.0.9000

*  ...
```

tests/testthat/test-scale.r
```diff
@@ -19,3 +19,24 @@ test_that("NA.value works for discrete", {
   expect_that(dscale(x, pal, "grey50")[1], equals("grey50"))
   expect_that(dscale(x, pal, "grey50")[5], equals("grey50"))
 })
+
+test_that("train_continuous maintains class", {
+  expect_equal(train_continuous(1:5),
+               c(1, 5),
+               info="Normal numeric input works with train_continuous")
```
dpseidel: Can you please style your tests according to [style.tidyverse.org](http://style.tidyverse.org)? There should be space around each equal sign. The styler package can help with this.

NEWS.md
billdenney: When this is resolved, I'll provide a new PR.

dpseidel: So no, this change will not be included in the 1.0.0 release. Please move this news bullet to the top with the new header I demonstrated above, and add your github handle and the issue number so that it follows [style guidelines](http://style.tidyverse.org/news.html#news-bullets). No need to open a new PR, just make a new commit to your branch with the changes.

billdenney:
I think that I got everything with this update.

NEWS.md
```diff
@@ -1,3 +1,10 @@
+# scales 1.0.0.9000
+
+## Minor bug fixes and improvements
```
dpseidel: This subheading is unnecessary when listing changes to the dev version, it's added when organizing changes just before release. Thanks for your patience with all these little nits -- you caught us right at release weeks so versions in NEWS and DESCRIPTION, haven't been bumped yet.

billdenney: No worries.  I'd be making similar comments on people submitting PRs to my repositories.

tests/testthat/test-scale.r
```diff
@@ -19,3 +19,24 @@ test_that("NA.value works for discrete", {
   expect_that(dscale(x, pal, "grey50")[1], equals("grey50"))
   expect_that(dscale(x, pal, "grey50")[5], equals("grey50"))
 })
+
+test_that("train_continuous maintains class", {
+  expect_equal(train_continuous(1:5),
+               c(1, 5),
+               info = "Normal numeric input works with train_continuous")
```
hadley: Please remove all uses of `info`

tests/testthat/test-scale.r
```diff
@@ -19,3 +19,24 @@ test_that("NA.value works for discrete", {
...
+  expect_equal(train_continuous(my_date),
+               c(my_date, my_date),
+               info = "Class is maintained through train_continuous with NULL existing")
+  expect_equal(train_continuous(my_date, existing = 1),
+               c(1, my_date),
+               info = "Class is changed through train_continuous with numeric existing")
+})
+
+test_that("train_continuous with NULL input", {
+  expect_equal(train_continuous(new = NULL, existing = c(1, 5)),
```
hadley: It seems unnecessary to name both arguments to `train_continuous()`

R/scale-continuous.r
```diff
@@ -38,7 +38,11 @@ train_continuous <- function(new, existing = NULL) {
   if (is.factor(new) || !typeof(new) %in% c("integer", "double")) {
     stop("Discrete value supplied to continuous scale", call. = FALSE)
   }
-  suppressWarnings(range(existing, new, na.rm = TRUE, finite = TRUE))
+  if (is.null(existing)) {
```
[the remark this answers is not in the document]
billdenney: I would have expected that, too, but it doesn't:

```r
> range(as.POSIXct("2018-01-31"), NULL)
[1] "2018-01-31 EST" "2018-01-31 EST"
> range(NULL, as.POSIXct("2018-01-31"))
[1] 1517374800 1517374800
> 
```

tests/testthat/test-scale.r
billdenney: I can do that, but why remove it?  I find it to be useful to understand the underlying rationale for the test.  Is there a preferred way to indicate rationale?

R/scale-continuous.r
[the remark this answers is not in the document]
hadley: Oh this is related to the type nightmare that is `c()`.

Can you add a comment to that affect? i.e. `range(NULL, x) strips attributes`

tests/testthat/test-scale.r
hadley: Rational should usually be encoded in the test name, or in a comment. In this case, I don't think that "Normal numeric input works with train_continuous" adds anything to my understanding of the test - it's implied by the test name that you'll be iterating through different types, and "working" is implied by `expect_equal()`

billdenney:
@hadley, I think that I've resolved everything.  There  is a ≥ symbol apparently showing up somewhere in the documentation causing the manual build to fail, but I think that's unrelated to this PR

tests/testthat/test-scale.r
```diff
@@ -19,3 +19,36 @@ test_that("NA.value works for discrete", {
   expect_that(dscale(x, pal, "grey50")[1], equals("grey50"))
   expect_that(dscale(x, pal, "grey50")[5], equals("grey50"))
 })
+
+test_that(paste(
```
hadley: Can you please split this into two tests?

tests/testthat/test-scale.r
```diff
@@ -19,3 +19,36 @@ test_that("NA.value works for discrete", {
...
+  "train_continuous maintains class with `existing=NULL`,",
+  "and changes class to the class of `existing` when not NULL."
+), {
+  expect_equal(
+    train_continuous(1:5),
+    c(1, 5)
+  )
+  expect_error(train_continuous(LETTERS[1:5]),
+    regexp = "Discrete value supplied to continuous scale"
```
hadley: This regexp is a bit long - the point is to check that you're got the right error, not to validate the exact text of the error message.

tests/testthat/test-scale.r
```diff
@@ -19,3 +19,36 @@ test_that("NA.value works for discrete", {
   expect_that(dscale(x, pal, "grey50")[1], equals("grey50"))
   expect_that(dscale(x, pal, "grey50")[5], equals("grey50"))
 })
+
+test_that(paste(
+  "train_continuous maintains class with `existing=NULL`,",
+  "and changes class to the class of `existing` when not NULL."
+), {
+  expect_equal(
```
hadley: I think these can now fit on one line

tests/testthat/test-scale.r
hadley: But why is it in this test? It doesn't seem to be related to the topic of test.

billdenney: Good point, I've separated it into another test (and while not technically related to this PR's main topic, it is related to increasing test coverage for `train_continuous()`).

hadley:
Thanks!

PR #169, opened 2018-08-09
Make log_breaks work with very small ranges

### LINKED ISSUE #168: log_breaks Not Always Useful
This is very similar to #100 but a more specialized case.

The second example below does not provide useful breaks (none show up when plotted).  I like the "relevant_breaks" addition in #117, and similar code should be added to `log_sub_breaks`.  If `log_sub_breaks` cannot find `>= (n-2)` breaks, then I think we should fall back to `extended_breaks` which has better guarantees of the number of relevant breaks even if we lose the equidistant log-scale spacing.

``` r
library(scales, quietly = TRUE)
(scales::log_breaks(base = exp(1)))(c(2, 2.0001))
#> [1] 1.000000 2.000000 2.718282
(scales::log_breaks(base = exp(1)))(c(2.0001, 2.0002))
#> [1] 2.000000 2.718282
(scales::extended_breaks())(c(2.0001, 2.0002))
#> [1] 2.000100 2.000125 2.000150 2.000175 2.000200
```

This affects many figures for me, so if possible to make it into 1.0.0, I think that would be selfishly beneficial (#157).

billdenney:
Fix #168

### DIFF
--- a/tests/testthat/test-breaks-log.r
+++ b/tests/testthat/test-breaks-log.r
@@ -64,14 +64,22 @@ test_that("log_breaks arguments are forcely evaluated on each call #81", {
 })
 
 test_that("log_breaks with very small ranges fall back to extended_breaks", {
-  expect_equal(log_breaks(n = 5, base = 10)(c(2.001, 2.002)),
-               extended_breaks(n = 5)(c(2.001, 2.002)))
-  expect_equal(log_breaks(n = 5, base = 10)(c(0.95, 1.1)),
-               extended_breaks(n = 5)(c(0.95, 1.1)))
+  expect_equal(
+    log_breaks(n = 5, base = 10)(c(2.001, 2.002)),
+    extended_breaks(n = 5)(c(2.001, 2.002))
+  )
+  expect_equal(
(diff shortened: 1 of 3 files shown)


billdenney:
I found and fixed one.  (Crosses fingers that was the only one.)

tests/testthat/test-breaks-log.r
```diff
@@ -62,3 +62,16 @@ test_that("log_breaks arguments are forcely evaluated on each call #81", {
   expect_equal(subfun1(1:1000), subfuns[[1]](1:1000))
   expect_equal(subfun2(1:1000), subfuns[[2]](1:1000))
 })
+
+test_that("log_breaks with very small ranges fall back to extended_breaks", {
+  expect_equal(log_breaks(n = 5, base = 10)(c(2.001, 2.002)),
```
hadley: Oops, I missed that these don't use the tidyverse indent style: http://style.tidyverse.org/syntax.html#long-lines

billdenney: I succumbed and installed the `styler` package.  :)

hadley:
Thanks!!

PR #170, opened 2018-08-09
Fix R6 inheritance

dpseidel:
In trying to covert ggplot2 to use scales exported `Range` objects (tidyverse/ggplot2#2710), I realized that the R6 inheritance was broken. This PR fixes that. Unfortunately this likely means that conversion in ggplot2 needs to wait until the next scales release.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # scales 1.0.0.9000
 
+* `ContinuousRange` and `DiscreteRange` methods now properly inherit and are fully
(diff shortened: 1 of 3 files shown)


hadley:
Could you please add a unit test too?

dpseidel:
Added! Good thing too! They helped me catch another bug -- test-based development for the win! 🏆

hadley:
Is probably now deserving of a news bullet.

PR #176, opened 2018-09-05
Enforce units for symbols in number_bytes() and use symbols = "auto" for showing varying multiples

### LINKED ISSUE #174: Bytes is with a capital B
The functions number_bytes_format() and number_bytes() incorrectly return a lower case "b". 

> number_bytes_format()(sample(3000000000, 3))
[1] "899 Mb" "541 Mb" "78 Mb" 

But that should read 
[1] "899 MB" "541 MB" "78 MB".

In fact, the web page cited in the help document (http://en.wikipedia.org/wiki/Units_of_information)  states that:
> The IEEE 1541-2002 standard specifies "B" (upper case) as the symbol for byte 

The lowercase "b" is for bits. I wouldn't quibble except that the function has the word "byte" in the name.  (Scales is an essential package. Thanks for writing it!)

mikmart:
This branch includes a few changes to `number_bytes()`:

1. Units for symbols are enforced, i.e. can't have `units = "si"` and `symbol = "MiB"` or `units = "binary"` and `symbol = "MB"`. (This also fixes #174 as `*b` isn't valid in either unit system.)
3. Zero bytes are formatted: `number_bytes(0) == "0 B"`. This previously resulted in `character(0)`.
4. **Changed the meaning of `symbol = "auto"`** to use varying multiples, i.e. `number_bytes(1024^(1:2)) == c("1 KiB", "1 MiB")`. This (to me at least) seems like a more useful "auto" feature than picking a fixed symbol based on the maximum value: e.g. when formatting axis labels, it's easy to find the largest symbol with a two-pass method and then give it as a fixed argument (plot once, pick largest), while varying multiples would require a wrapper function with `map_chr()` or equivalent.

Do these changes sound reasonable? I haven't updated any of the docs (yet), but I'll be happy to add that if this were to be implemented.


R/formatter.r
```diff
@@ -767,49 +767,32 @@ number_bytes_format <- function(symbol = "auto", units = "binary", ...) {
...
+      si     = c( "kB",  "MB",  "GB",  "TB",  "PB",  "EB",  "ZB",  "YB"),
+      binary = c("KiB", "MiB", "GiB", "TiB", "PiB", "EiB", "ZiB", "YiB")
+    )
   )
 
   if (!(symbol %in% symbols)) {
-    warning(paste(
-      "`symbol` must be one of '", paste0(symbols, collapse = "', '"),
-      "'. Defaulting to 'auto'."
+    warning(paste0(
```
hadley: Can you eliminate the `paste()` altogether? `warning()` already pastes together its inputs

R/formatter.r
```diff
@@ -767,49 +767,32 @@ number_bytes_format <- function(symbol = "auto", units = "binary", ...) {
...
+  units <- match.arg(units, c("binary", "si"))
+
   symbols <- c(
-    "auto",
-    "b", "Kb", "Mb", "Gb", "Tb", "Pb", "Eb", "Zb", "Yb",
-    "B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB",
-    "KiB", "MiB", "GiB", "TiB", "PiB", "EiB", "ZiB", "YiB"
+    "auto", "B",
+    switch(units,
+      si     = c( "kB",  "MB",  "GB",  "TB",  "PB",  "EB",  "ZB",  "YB"),
```
hadley: I wonder if it would be better to generate these from a data structure like `c(k = 3, M = 6, G = 9, ...)`, which could then be reused when automatically finding the symbol.

mikmart: @hadley yes I think that would be neater. I was thinking maybe it would make sense to just include a metric prefix table as package data and use it here? E.g. from [here](https://en.wikipedia.org/wiki/Metric_prefix)? Could include powers, symbols and text names for prefixes. It could also be used with `number_si()` and in other places, too, I'm sure.

mikmart: I can open another PR for that if that sounds good.

mikmart: @hadley I added the SI prefix table in #177, and it's working; but now I'm thinking maybe this got a bit over-engineered... Maybe it would be simpler to just include a vector like you suggested. What do you think?

hadley: Yeah, I think a vector would be better, although embedding the data frame directly isn't too bad if you think the other columns would be useful in the future.

mikmart: I think having a single place for all the SI prefixes would be useful, but perhaps not so much the columns other than power and symbol. So I went with a middle-ground and included a named vector outside the function.

R/formatter.r
```diff
@@ -740,18 +740,15 @@ pvalue <- function(x, accuracy = .001, decimal.mark = ".", add_p = FALSE) {
 
 #' Bytes formatter: convert to byte measurement and display symbol.
 #'
-#' @return a function with three parameters, `x``, a numeric vector that
-#'   returns a character vector, `symbol` the byte symbol (e.g. "Kb")
-#'   desired and the measurement `units` (traditional `binary` or
-#'   `si` for ISI metric units).
+#' @return a function with three parameters, `x``, a numeric vector that returns
```
hadley: Can you please fix the ``` `` ``` ? (it's not your fault, but you've touched this line so you might as well fix it)

R/formatter.r
```diff
@@ -760,56 +757,40 @@ pvalue <- function(x, accuracy = .001, decimal.mark = ".", add_p = FALSE) {
...
+    si     = paste0(prefix, "B"),
+    binary = paste0(toupper(prefix), "iB")
+  ))
 
   if (!(symbol %in% symbols)) {
-    warning(paste(
-      "`symbol` must be one of '", paste0(symbols, collapse = "', '"),
-      "'. Defaulting to 'auto'."
-    ), call. = F)
+    warning("`symbol` must be one of: '", paste0(symbols, collapse = "', '"),
```
hadley: Can you please follow the long-lines style: http://style.tidyverse.org/syntax.html#long-lines

hadley:

mikmart:
@hadley I added the bullet and addressed the other comments, too. This should be good to go.

hadley:
Thanks!!

PR #177, opened 2018-09-05
Include a table of SI prefixes as package data

mikmart:
Brought up in comments in #176 , this branch includes a table of SI prefixes as internal data for easy access to SI prefixes, their written names and corresponding powers.


hadley:
I don't think this needs to go in `sysdata.Rda` — you can just put `data-raw/sysdata.R` in `R/si-prefixes.R` (and don't document it since it's not exported)

mikmart:
I ended up reducing the table to a named vector instead (dropping the "text" column), and it's now included in `R/si-powers.R`.

mikmart:
Since #176 was merged I don't think this PR would actually do anything anymore, at least right now.

PR #178, opened 2018-09-06
Implement degree_format() (#159)

### LINKED ISSUE #159: Add a formatter for labels requiring a degree symbol
Add a formatter that adds degree symbols and characters to a numeric vector to make labels for things like temperature or latitude and longitude (as currently done by geom_sf). 

The current way to get this to format correctly in ggplot is something like:

```
library(ggplot2)
ggplot(data.frame(), aes(1, 1)) + scale_x_continuous(labels = c(
  expr(160 * degree * W), expr(140 * degree * W),
  expr(120 * degree * W), expr(100 * degree * W), 
  expr(0 * degree * W))
```

mikmart:
Implements `degree_format()` as a tweaked `unit_format()` that adds `°` (i.e. U+00B0) between the `sep` and `unit` in the `suffix`, and defaults to `sep = "", unit = ""`. Closes #159.

I didn't add a `degree()` as after a quick search that seems to be a fairly common function name, and I would imagine the default of no unit is not what most users would want anyway.


hadley:
Thanks!

PR #188, opened 2018-11-02
Allow formatting functions to skip NAs. Closes #187.

### LINKED ISSUE #187: Should format functions turn NAs into strings?
I just noticed that format functions turn `NA`s into strings instead of keeping them as `NA`. Is this as intended? It's particularly weird for the `dollar()` function.

``` r
x <- c(100, NA, 20000, NA, 1.5)
scales::comma(x)
#> [1] "100"    "NA"     "20,000" "NA"     "2"
scales::dollar(x)
#> [1] "$100.00"    "$NA"        "$20,000.00" "$NA"        "$1.50"
```

<sup>Created on 2018-11-01 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
This is a first stab at the `NA` issue in formatting functions. It's a bit tricky, because there are many different formatters, and for some it may be desired to turn `NA` into `"NA"` and for others it's clearly not desired. For the formatters that I have changed, I have added a toggle, but the default is now to keep `NA` as is.

There are three ways this could be done:
1. Change the default to not formatting `NA`s, as in this PR.
2. Keep the default as it is in the current code, even for functions where the current default is clearly a bad choice.
3. Make a decision on a case-by-case basis about what is the more appropriate default for each function.

I see downsides for each of these choices, so I'm not entirely sure what is the best way to proceed, and I'd like to know whether I'm approximately on the right track.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 * New `degree_format()` adds degree symbols to numbers (@mikmart, #159).
 
+* Formatting functions now consistently keep `NA`s as `NA`s instead of trying
+  to format them (@clauswilke, #187).
+
 * `number_bytes()` now:
 
     1. Always uses "B" as the symbol for bytes (#174).
--- a/R/formatter.r
+++ b/R/formatter.r
@@ -753,8 +753,7 @@ pvalue <- function(x, accuracy = .001, decimal.mark = ".", add_p = FALSE) {
     x,
     accuracy = accuracy,
     decimal.mark = decimal.mark,
-    big.mark = "",
-    na.exclude = FALSE
+    big.mark = ""
   )
   if (add_p) res <- paste0("p=", res)
   below <- number(
--- a/R/formatter.r
+++ b/R/formatter.r
@@ -32,8 +32,6 @@
 #'   decimal point.
 #' @param trim Logical, if `FALSE`, values are right-justified to a common
 #'   width (see [base::format()]).
-#' @param na.exclude Logical, if `TRUE` (the default), `NA` values are excluded
-#'   from formatting and kept as `NA`.
 #' @param ... Other arguments passed on to [base::format()].
 #' @export
 #' @examples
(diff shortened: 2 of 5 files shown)


hadley:
I am reasonably certain that formatting functions should always preserve `NA` as `NA` not translate to `"NA"`.

hadley:
And I'd slightly prefer modifying "in place" rather than using `ifelse()` (which has complex semantics), e.g. https://vctrs.r-lib.org/articles/s3-vector.html#format-method

hadley:
The cost of making it an option seems low, but adds additional complexity to every function that then needs to be carefully tested. I'd rather just do something sensible without an additional option (which would mean `pvalue()` would work slightly differently to the others, but I don't think that's a problem)

clauswilke:
I have committed a new version. It's much simpler and cleaner, and I think it supports your perspective of how this should be approached. The `pvalue()` function now automatically behaves as one would expect, without requiring any additional logic.

``` r
scales::pvalue(c(.5, .045, .0002, NA), accuracy = .01)
#> [1] "0.50"  "0.04"  "<0.01" NA
scales::pvalue(c(.5, .045, .0002, NA), add_p = TRUE)
#> [1] "p=0.500" "p=0.045" "p<0.001" "p=NA"
```

Should I tackle the other formatting functions also, scientific, date, ordinal, etc?

hadley:
That would be wonderful!

clauswilke:
I think I've covered all formatting functions. I also found a few other `NA`-related bugs that I fixed. Specifically, `percent(NA_real_)` doesn't work in current scales and `ordinal(c(21, 1, NA, 11))` produces incorrect output.

hadley:
@clauswilke is there other stuff you want to do with scales? If so, I can add to you to the scales admin. If not, I can merge this.

clauswilke:
@hadley Your call. I don't have any immediate plans right now, but because of the close relationship between scales and ggplot2 things may continue to come up. Curiously, this particular problem arose not in a plotting context but when I wanted to use `dollar()` to format currency values for a csv file.

hadley:
Ok, gave you access — in an ideal world this would be automatic when you got ggplot2 access; this is an unanticipated downside of putting it in a different org.

PR #189, opened 2018-11-06
Add power_of_ten transformation

Anirudhsekar96:
I have added a base functions to transform the scale of values as a power of ten. I have also included further derived functions to easily convert into thousands scale, millions scale, billions scale and trillions scale.  This functionality using the base function could be used to extend the derived functions to convert any arbitrary power of ten.

### DIFF
--- a/R/trans-numeric.r
+++ b/R/trans-numeric.r
@@ -208,3 +208,32 @@ pseudo_log_trans <- function(sigma = 1, base = exp(1)) {
     function(x) 2 * sigma * sinh(x * log(base))
   )
 }
+
+
+#' Power_of_ten transformation.
+#'
+#' @param base Power of ten
+#' @aliases power_of_ten_trans thousands_trans millions_trans billions_trans trillions_trans
+#' @export power_of_ten_trans thousands_trans millions_trans billions_trans trillions_trans
+power_of_ten_trans <- function(base = 1) {
+  force(base)
+  trans <- function(x) x/(10^base)
+  inv <- function(x) x*(10^base)
+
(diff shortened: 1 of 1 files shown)


dpseidel:
Thanks for opening a PR! I've changed the name only to be more concise. Could you please add some tests, a [news bullet](https://style.tidyverse.org/news.html), and style your code according to the [tidyverse style guide](https://style.tidyverse.org/)?  The `styler` package can help make this easier. Once you're done, you will also need to rebuild the documentation. 

@hadley do you think this is an appropriate addition to the package?

hadley:
Don’t we already have a function that does something similar? (With a different parameterisation)

dpseidel:
As far as I can tell, we don't have an existing transformation that does this. Within the number formatters, we have several that do something like this for labels, which may solve most use cases rather than requiring a full transformation. Most simply, `number` and `number_format` can transform labels simply by setting the `scale` argument:

``` r
library(scales)
test <- c(1000, 20e3, 50e5)
number(test, scale = 1/1000)
#> [1] "1"     "20"    "5 000"
```

Additionally we recently added the `number_si` function, available in the dev version,  which has an auto-scaling functionality:

```r
number_si(test)
#>[1] "1K"  "20K" "5M" 
```
<sup>Created on 2019-04-30 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

hadley:
@Anirudhsekar96 I think this is a bit too duplicative of existing functions, especially given that I think `number_si()` gives nicer labels. But thanks for offering your code and I really appreciate the effort 😁

PR #191, opened 2018-11-14
Port improvements to color mapping functions from leaflet

jcheng5:
* `col_numeric()`, `col_bin()`, `col_quantile()`, and `col_factor()` now support
  viridis colors. Just pass a palette name (`"magma"`, `"inferno"`, `"plasma"`,
  or `"viridis"`) as the `palette` argument.

* `col_numeric()`, `col_bin()`, `col_quantile()`, and `col_factor()` now have a
  `reverse` parameter, to apply color palettes in the opposite of their usual
  order (i.e. high-to-low instead of low-to-high).

* `col_bin()` and `col_quantile()` now take a `right` argument, which is passed
  to `base::cut()`; it indicates whether the bin/quantile intervals should be
  closed on the right (and open on the left), or vice versa.

* `col_factor()` now tries to avoid interpolating qualitative RColorBrewer
  palettes. Instead, it attempts to assign a palette color to each factor level.
  Interpolation will still be used if there are more factor levels than
  available colors, and a warning will be emitted in that case.


jcheng5:
@hadley I've implemented your suggestions. There's one significant difference in behavior between my `brewer_pal_colors` and `brewer_pal`, that occurs when you request 2 colors from an RColorBrewer palette (such as when coloring a factor with only 2 levels). The `RColorBrewer::brewer.pal` call will return three colors; `brewer_pal_colors` will use the first and third, while `brewer_pal` will use the first and second.

Here's a demonstration of the differences:
http://rpubs.com/jcheng/small-palettes

I thought I'd feel more strongly about the leaflet version being superior, but now looking at that link, I'm not sure it's that big of a deal either way. I do think the Sequential/Blues looks better for leaflet, and Qualitative/Accent looks better for scales.

hadley:
I don't feel particularly strongly, but I do think return the first two colours is more consistent that returning the first and the third.

jcheng5:
@hadley I'm fine with that. Let me know if you see anything else I need to change--otherwise I think it's done.

R/colour-mapping.r
```diff
@@ -261,42 +278,51 @@ col_factor <- function(palette, domain, levels = NULL, ordered = FALSE,
...
+      colors <- brewer_pal(palette = pal)(RColorBrewer::brewer.pal.info[pal, "maxcolors"]) # Get all colors
+    }
+  } else if (length(pal) == 1 && pal %in% c("viridis", "magma", "inferno", "plasma")) {
+    colors <- viridis_pal(option = pal)(256)
+  } else {
+    colors <- pal
   }
 
-  colour_ramp(pal)
+  scales::colour_ramp(colors, alpha = alpha)
```
hadley: Just noticed this, which can now be removed. Otherwise, looks good.

PR #195, opened 2019-01-11
preserve element names in `alpha()`

wibeasley:
The `alpha()` function now preserves names on the elements.  I followed the approach used earlier in the file for `col2hcl()`.

https://github.com/r-lib/scales/blob/51547d6865f61f8cc0239063bacc619dce8a35b6/R/colour-manip.r#L26

A common use case for me is when I create palette to pass to the `values` parameter of `ggplot2::scales_color_manual()` and `ggplot2::scales_fill_manual()`.  Like the [chapter 5 scatterplots here](
https://github.com/OuhscBbmc/DeSheaToothakerIntroStats/blob/master/thumbnails/thumbnails.md) (and  countless people before me), the scatterpoints are darker around the edges, but translucent in the center.  I need explicit vector names to force colors to be consistent across plots, even if a factor level is missing.

image

Instead of replicating the rgb components like:
```r
palette_dark  <- c("FALSE"="#3288BD"  , "TRUE"="#D53E4F"  ) # http://colrd.com/palette/18893/
palette_light <- c("FALSE"="#3288BDAA", "TRUE"="#D53E4FAA") 
```

It would be nice to reuse the palette (especially if wasn't defined immediately above):
```r
palette_dark  <- c("FALSE"="#3288BD", "TRUE"="#D53E4F") 
palette_light <- scales::alpha(palette_dark, .5)
```

### DIFF
--- a/R/colour-manip.r
+++ b/R/colour-manip.r
@@ -50,8 +50,6 @@ muted <- function(colour, l=30, c=70) col2hcl(colour, l = l, c = c)
 #' alpha("red", 0.1)
 #' alpha(colours(), 0.5)
 #' alpha("red", seq(0, 1, length.out = 10))
-#' alpha(c(`1` = "gold", `2` = "lightgray", `3` = "#cd7f32"), .5)
-#' alpha(c(first = "gold", second = "lightgray", third = "#cd7f32"), .5)
 #' alpha(c("first" = "gold", "second" = "lightgray", "third" = "#cd7f32"), .5)
 alpha <- function(colour, alpha = NA) {
   col <- grDevices::col2rgb(colour, TRUE) / 255
--- a/R/colour-manip.r
+++ b/R/colour-manip.r
@@ -50,9 +50,9 @@ muted <- function(colour, l=30, c=70) col2hcl(colour, l = l, c = c)
 #' alpha("red", 0.1)
 #' alpha(colours(), 0.5)
 #' alpha("red", seq(0, 1, length.out = 10))
-#' alpha(c(`1`="gold", `2`="lightgray", `3`="#cd7f32"), .5)
-#' alpha(c(first="gold", second="lightgray", third="#cd7f32"), .5)
-#' alpha(c("first"="gold", "second"="lightgray", "third"="#cd7f32"), .5)
+#' alpha(c(`1` = "gold", `2` = "lightgray", `3` = "#cd7f32"), .5)
+#' alpha(c(first = "gold", second = "lightgray", third = "#cd7f32"), .5)
+#' alpha(c("first" = "gold", "second" = "lightgray", "third" = "#cd7f32"), .5)
 alpha <- function(colour, alpha = NA) {
   col <- grDevices::col2rgb(colour, TRUE) / 255
 
--- a/R/colour-manip.r
+++ b/R/colour-manip.r
@@ -50,6 +50,9 @@ muted <- function(colour, l=30, c=70) col2hcl(colour, l = l, c = c)
 #' alpha("red", 0.1)
 #' alpha(colours(), 0.5)
 #' alpha("red", seq(0, 1, length.out = 10))
+#' alpha(c(`1`="gold", `2`="lightgray", `3`="#cd7f32"), .5)
+#' alpha(c(first="gold", second="lightgray", third="#cd7f32"), .5)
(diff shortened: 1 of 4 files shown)


wibeasley:
I see Travis failed with warnings about LaTeX and encoding.  I assume that's unrelated to my additions to the Rd file.  (I don't get those warnings on my local Ubuntu machine.)  But tell me if there's something I should modify.
```
checking PDF version of manual without hyperrefs or index ... ERRORchecking PDF version of manual ... WARNING
LaTeX errors when creating PDF version.
This typically indicates Rd problems.
LaTeX errors found:
! Package inputenc Error: Unicode char ≥ (U+2265)
(inputenc)                not set up for use with LaTeX.
```

dpseidel:
The build failure seems to be triggered by the `≥` symbols in the `number_si()` function documentation added over 5 months ago. We haven't caught it before since it appears in previous builds travis included a `-no-manual` flag to the R CMD check call. Similarly, I can reproduce this on my Mac only if I run `devtools::check(manual = T)`. 

@hadley or @jimhester is there something I need to add to the travis.yml to fix the latex install and this problem? Alternatively we could remove or escape these special characters?

@wibeasley Thanks for your PR! Would you mind tidy styling the code in your examples and tests? scales adheres to the [tidyverse style guide](https://style.tidyverse.org). Additionally, can you please add a bullet to NEWS.md? It should briefly describe the change and end with (@yourname, #issuenumber).

R/colour-manip.r
```diff
@@ -50,6 +50,9 @@ muted <- function(colour, l=30, c=70) col2hcl(colour, l = l, c = c)
 #' alpha("red", 0.1)
 #' alpha(colours(), 0.5)
 #' alpha("red", seq(0, 1, length.out = 10))
+#' alpha(c(`1` = "gold", `2` = "lightgray", `3` = "#cd7f32"), .5)
```
hadley: I think you only need a single example here

tests/testthat/test-alpha.r
```diff
@@ -29,3 +29,22 @@ test_that("col values recycled to match alpha", {
 
   expect_equal(alphas, reds_alpha)
 })
+
+test_that("preserves element names", {
+  expect_equal(c("first", "second", "third"), names(alpha(c(first   = "gold", second   = "lightgray", third   = "#cd7f32"), .5)))   # not quoted
```
hadley: I think you only need a single expectation here; otherwise you're acutally testing the behaviour of `names<-`.

R/colour-manip.r
wibeasley: @hadley, I thought it would be helpful for the example to show (a) backticks, (b) quoted, and (c) unquoted. 

Does that change your mind?  If you still think the triplicate is tedious, I'll remove two.  Which line(s) are the preferred tidyverse style?  

I think I'd prefer quotes here, because sometimes these levels are assigned dynamically inside a function.

tests/testthat/test-alpha.r
wibeasley: @hadley, I'm leaving the two cases where the first/last element is unnamed.  It's a habit of mine protecting against one-off errors.  If you feel strongly, I'll thin it out more.

R/colour-manip.r
hadley: I think `c("a" = "b")` is the best style. There's no need to elucidate every feature of `c()` here.

tests/testthat/test-alpha.r
```diff
@@ -29,3 +29,18 @@ test_that("col values recycled to match alpha", {
 
   expect_equal(alphas, reds_alpha)
 })
+
+test_that("preserves element names", {
+  expect_equal(c("first", "second", "third"), names(alpha(c("first" = "gold", "second" = "lightgray", "third" = "#cd7f32"), .5)))   # quoted
```
hadley: Again, given that all the code does is call `names<-`, I don't think these tests are necessary. You are adding additional overhead (tests that need to be read and understood) for no additional robustness.

hadley:
Thanks! For future reference, if you include `Fixes #xyz` in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

wibeasley:
Cool.  I didn't realize there was an issue for this outside my PR.  I'll keep an eye out in the future.

Issue #196, opened 2019-01-25
Yeo-Johnson Transform


zamorarr:
Hi `scales` devs - thanks for the great package. I would like to propose adding the [Yeo-Johnson](https://en.wikipedia.org/wiki/Power_transform#Yeo%E2%80%93Johnson_transformation) transform to complement the existing BoxCox and Modulus transforms. The Yeo-Johnson transform can be useful for pre-processing data and is already included in the `recipes` and `caret` packages. It would be great to  have it in this package to make visualization easy in `ggplot2`.

I'd be happy to submit a PR for review.

dpseidel:
I'm happy to review a PR for this. @hadley do you think this is an appropriate addition?

hadley:
Sure, seems reasonable.

PR #198, opened 2019-01-28
Yeo-Johnson Transform

zamorarr:
This PR addresses #196 and implements the [Yeo-Johnson](https://en.wikipedia.org/wiki/Power_transform#Yeo%E2%80%93Johnson_transformation) transform for continuous values. The Yeo-Johnson transform has 4 separate forms depending on the values of lambda and x:

- x >= 0, lambda == 0
- x >= 0, lambda != 0
- x < 0, lambda == 2
- x < 0 , lambda != 2

I broke the transformation up into two inner functions, `trans_pos` (for x >= 0 cases) and `trans_neg` for (x < 0 cases). I also included a `cutoff` value to provide some tolerance when checking if lambda is equal to 0 or 2, but I chose the value arbitrarily. 

Appreciate your review @dpseidel !


zamorarr:
Hmm.. Looks like there are some travis errors related to building the PDF manual. I think I found the problem but it's not related to my new code so I'll submit a separate bugfix PR.

hadley:
Thanks @zamorarr!

zamorarr:
thanks for accepting!

PR #199, opened 2019-01-28
latex doesn't like unicode character for >=

zamorarr:
There seems to be an error related to building the PDF manual causing the following PRs to fail:

-  #198 
- #197 
- #195 

The error is:
```
LaTeX errors found:
! Package inputenc Error: Unicode char ≥ (U+2265)
(inputenc)                not set up for use with LaTeX.
```

which is caused by the use of the `≥` symbol in the docs for `number_format()`. I've changed the offending symbol.


dpseidel:
Great! Thank you!

Issue #203, opened 2019-04-09
Parse problem for label axis using scales::parse_format


agila5:
Hi! I want to create a plot with col and axis labels that need to be firstly parsed. I read [here](https://github.com/tidyverse/ggplot2/issues/311) that I have to use scales::parse_format and, while it works for the fill legend, it does not seem to work for the x axis since it shows only the ticks . Am I doing something wrong? Is this a bug? 

I originally posted [the same](https://github.com/tidyverse/ggplot2/issues/3228) issue at the github page of ggplot2 and they suggested that this could be a bug with scales since the same example works using ggplot2 internal functions. 

``` r
# packages
library(ggplot2)

# data
fake_data <- data.frame(
  classes = c("alpha", "beta"),
  value = c(1, 2),
  stringsAsFactors = FALSE
)

ggplot(fake_data) + 
  geom_col(aes(x = classes, y = value, fill = classes)) + 
  scale_fill_discrete(labels = scales::parse_format()) + 
  scale_x_discrete(labels = scales::parse_format())
```

![](https://i.imgur.com/DTwvupv.png)

<sup>Created on 2019-04-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

clauswilke:
The problem is that `scales::parse_format()` produces a list of expressions when what is needed is an expression object:
``` r
scales::parse_format()(c("alpha", "beta"))
#> [[1]]
#> expression(alpha)
#> 
#> [[2]]
#> expression(beta)
ggplot2:::parse_safe(c("alpha", "beta"))
#> expression(alpha, beta)
```

<sup>Created on 2019-04-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

This problem was discussed in detail in this ggplot2 issue: https://github.com/tidyverse/ggplot2/issues/2864

Probably the right way forward is to copy `ggplot2:::parse_safe()` over to scales and use it for `scales::parse_format()`.

dpseidel:
Interesting to me that the list of expressions works for one discrete scale but not the other....:thinking: 

I can build a PR and test this change out later this week (dissertation deadlines looming right now). Or, alternatively, happy to help review if either of you want to work one up a little sooner.

clauswilke:
> Interesting to me that the list of expressions works for one discrete scale but not the other....🤔
[...]

PR #204, opened 2019-04-11
change parse_format function

agila5:
I change the `parse_format` function according to what was discussed in the issue https://github.com/r-lib/scales/issues/203. Now the same code retuns:

``` r
library(ggplot2)

# data
fake_data <- data.frame(
  classes = c("alpha", "beta"),
  value = c(1, 2),
  stringsAsFactors = FALSE
)

ggplot(fake_data) + 
  geom_col(aes(x = classes, y = value, fill = classes)) + 
  scale_fill_discrete(labels = scales::parse_format) + 
  scale_x_discrete(labels = scales::parse_format)
```

<sup>Created on 2019-04-11 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

The strange thing is that the same code returns a different plot (i.e. the fill labels are too close to the colours) if I view it using Rstudio: 
Rplot


R/formatter.r
```diff
@@ -588,18 +588,28 @@ ordinal_spanish <- function() {
...
+#'
 #' @export
 #' @examples
-#' parse_format()(c("alpha", "beta", "gamma"))
-parse_format <- function() {
-  function(x) {
-    lapply(as.character(x), function(x) parse(text = x, srcfile = NULL))
+#' parse_format(c("alpha", "beta", "gamma"))
+#' parse_format(1:5)
+parse_format <- function(text) {
```
hadley: You need to preserve the original structure of this function you code should look like:

```R
parse_format <- function() {
  function(x) {
     x <- as.character(x) 
     out <- ...
     ...
  }
}
```

hadley:
Thanks!

For future reference, if you include `Fixes #xyz` in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

agila5:
> For future reference, if you include Fixes #xyz in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

Ok, thanks. I always forget about that. 

Anyway, can you reproduce the problem with the labels positioning in the legend? I still get the same plot as in the first post of this conversation where, if I view the plot using Rstudio, the labels, in my opinion, are way too close to the legend. If I use the CRAN version of `scales` (i.e. the old version of `parse_format `), the problem disappears.

hadley:
The code that works for me is

```R
ggplot(fake_data) + 
  geom_col(aes(x = classes, y = value, fill = classes)) + 
  scale_fill_discrete(labels = scales::parse_format()) + 
  scale_x_discrete(labels = scales::parse_format())
```

But if you're having a problem, it would be better to file a new issue, rather than discussing it in a closed PR.

PR #211, opened 2019-06-16
Update index.html to reflect ver. 1.0.0 release

tdawry:
Previously showed scales 1.0.0 as "Unreleased"


hadley:
This is generated automatically so patching the html file isn't quite the right solution. The website looks correct now, so I assume something went wrong with the automated build that's now been fixed.

PR #220, opened 2019-09-04
Add exponent and exact_zero argument to scientific_format

teramonagi:
## Summary
The notation with 10 as a power index is [often discussed](https://stackoverflow.com/questions/10762287/how-can-i-format-axis-labels-with-exponents-with-ggplot2-and-scales).
Therefore, I thought it is useful to include this notation in `scales` package and implemented it.

## Example
```r
> scientific(1:10)
 [1] "1e+00" "2e+00" "3e+00" "4e+00" "5e+00" "6e+00" "7e+00" "8e+00" "9e+00" "1e+01"

> scientific(0:10, exponent="10", exact_zero = FALSE)
expression(0 %*% 10^00, 1 %*% 10^00, 2 %*% 10^00, 3 %*% 10^00, 
    4 %*% 10^00, 5 %*% 10^00, 6 %*% 10^00, 7 %*% 10^00, 8 %*% 10^00, 
    9 %*% 10^00, 1 %*% 10^01)

> scientific(0:10, exponent="10", exact_zero = TRUE)
expression(0, 1 %*% 10^00, 2 %*% 10^00, 3 %*% 10^00, 4 %*% 10^00, 
    5 %*% 10^00, 6 %*% 10^00, 7 %*% 10^00, 8 %*% 10^00, 9 %*% 10^00, 
    1 %*% 10^01)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ sudo: false
 cache: packages
 
 r:
-- 3.1
 - 3.2
 - 3.3
 - oldrel
--- a/R/formatter.r
+++ b/R/formatter.r
@@ -432,17 +432,19 @@ dollar <- function(x, accuracy = NULL, scale = 1, prefix = "$",
 
 #' Scientific formatter.
 #'
-#' @return A function with single parameter `x`, a numeric vector, that
+#' @return A function with single parameter x, a numeric vector, that
 #'   returns a character vector.
 #' @param digits Number of significant digits to show.
-#' @param scale A scaling factor: `x` will be multiply by `scale` before
+#' @param scale A scaling factor: x will be multiply by scale before
 #'   formating (useful if the underlying data is on another scale,
 #'   e.g. for computing percentages or thousands).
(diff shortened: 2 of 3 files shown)


teramonagi:
The error from travis CI is 

> Error: (converted from warning) packages ‘httr’, ‘rlang’, ‘vctrs’ are not available (for R version 3.1.3)
Exe

.
This error occurs only on R 3.1. 
Do we need to take care R version 3.1?

yutannihilation:
> Do we need to take care R version 3.1?

No. tidyverse supports only the following versions, and I believe the same policy applies on the r-lib packages (so we see the error that complains httr, rlang, and vctrs are not available on R 3.1.x).

> the current version, the devel version, and four previous versions of R.
(https://www.tidyverse.org/articles/2019/04/r-version-support/)

teramonagi:
Thank you for your advice.
I removed R 3.1 support from `.travis.yml`.
If I do in the other PR, let me know.

PR #221, opened 2019-09-24
BUG FIX: Missing importFrom(utils,globalVariables)

HenrikBengtsson:
# Issue
`globalVariables()` is used without being imported from 'utils'.

# Background

I was getting an `devtools::check()`/`R CMD check` error on:
```r
checking whether the package can be loaded with stated dependencies ...Loading required package: TopDom
   Error in globalVariables(".x") : 
     could not find function "globalVariables"
   Error: package or namespace load failed for ‘TopDom’:
    unable to load R code in package ‘scales’
   Error: package ‘TopDom’ could not be loaded
   Execution halted
```
when testing an in-house package where 'scales' is a nested dependency somewhere deep down.  This PR fixed the problem.


hadley:
Thanks — I fixed a slightly different way

PR #224, opened 2019-10-25
Use farver for colour transformations

### LINKED ISSUE #223: Use farver for colour ramp
Instead of home grown (?) C++ code

hadley:
Fixes #223

* [x] Update all tests with exact colours
* [x] Replace `convertColor()` calls
* [x] Rewrite `col2hcl()`
* [x] Replace all other uses of grDevices
* [x] Create benchmark for `color_ramp()`  vs `colorRamp()`


hadley:
Unfortunately this change makes `colour_ramp()` slower than `colorRamp()`, at least in R3.6:

``` r
x <- seq(0, 1, length = 1e5)
col <- c("red", "green", "blue")

bench::mark(
  base = grDevices::rgb(colorRamp(col)(x), maxColorValue = 255)
)
#> # A tibble: 1 x 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 base         17.2ms   21.9ms      45.3    19.4MB     34.0
bench::mark(
  scales = scales::colour_ramp(col)(x)
)
#> # A tibble: 1 x 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 scales       49.5ms   51.4ms      19.4    36.7MB     58.3
```

<sup>Created on 2019-10-31 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I did a little profiling and the bottleneck is primarily in the matrix manip in `col2rgb()` and `rgb2col()` so I'm confident the performance will improve drastically once those functions are implemented natively in farver.

hadley:
Oh it's also slower in R 3.5 with cran scales, so I don't think this is a new change, and I don't have any hesitation to merge this PR as is.

hadley:
@thomasp85 I don't think you need to prioritise it; if it happens, it happens.

PR #235, opened 2019-11-20
Fix label_number_si() to use SI prefixes

### LINKED ISSUE #234: label_number_si does not use SI prefixes (Billion vs Giga)
Thanks for this release. It looks like it delivers great improvements.

The `label_number_si()` uses [short scale (English) prefixes](https://en.m.wikipedia.org/wiki/Long_and_short_scales) instead of [SI prefixes](https://en.m.wikipedia.org/wiki/Metric_prefix). The naming is confusing, since with that function name I would expect labels like "1k, 1M, 1G".

In several European countries 1 billion is 1 million millions. As far as I know there is no SI agreement on what a billion is, and the SI recommends using metric prefixes (1G for 1E9).

Also, the documentation for that function states 10e9, when it means 1e9 ("e" means "multiplied by 10 to the power of").

Either there is a rationale I am missing or without a breaking change this behaviour will last for a long time... If that's the case I hope there is an early bugfix release that breaks backwards compatibility...

Thanks again!

### LINKED ISSUE #301: Possibly confusing default for label_number_si?
I've found the relatively new `label_number_si()` quite useful -- thanks! One thing that just caught me surprised is how it handles going from non SI to SI units. For example, at first glance, the axis below looks correct. But in fact, we get 2K because 1.5K is being rounded up as the default accuracy is 1.

``` r
library(scales)

demo_continuous(c(0, 1500), label = label_number_si())
#> scale_x_continuous(label = label_number_si())
```

Taking the accuracy to 0.1 works, but then you get the unnecessary decimal places at 0 and 500.

``` r
demo_continuous(c(0, 1500), label = label_number_si(0.1))
#> scale_x_continuous(label = label_number_si(0.1))
```

<sup>Created on 2021-01-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

davidchall:
Fixes #234. Fixes #301.

**TLDR:** `label_number_si()` is currently providing [short scale abbreviations](https://en.wikipedia.org/wiki/Long_and_short_scales) instead of [SI prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes).

Both SI prefixes and long/short scale abbreviations are solutions to the same problem – how to specify large values more concisely. SI prefixes are now favored in most domains except finance. Abbreviations of long/short scales are not standardized (e.g. million could be "M", "m", "mn" in different circles) and depends on locale.

I've done my best to adjust the interface, but would love to hear feedback on how I could improve!

Here's what changed:
- Changes to `label_number_si()`
  - Now uses SI prefixes
  - The `unit` argument was promoted to a required argument
  - Added `scale` argument
  - Default `accuracy` changed to `NULL` (i.e. auto precision)
  - Removed `sep` argument (wasn't actually doing anything)
- Changes to `label_dollar()`
  - Added `rescale_large` argument
  - Added `rescale_short_scale()` and `rescale_long_scale()` built-in values, to use with this argument
- Changes to `label_bytes()`
  - Handle `scale` argument (would've been passed by `...`)
- Added examples and tests


davidchall:
The Windows CI is failing because one of the SI prefixes is a unicode character (μ), and the test appears unable to compare these correctly. I've not used R on Windows before, so I think I'll need some help with this.

I also did a bit of reading about [long and short scales](https://en.wikipedia.org/wiki/Long_and_short_scales) (the previous implementation of this function) and it seems that these are now mostly only used with currencies. Perhaps it makes more sense to include this previous functionality as an optional argument to `label_dollar()`? E.g.

```r
test that::expect_equal(
    label_dollar(c(1, 1e3, 1e6, 1e9, 1e12), billion_scale = "short"),
    c("$1", "$1k", "$1m", "$1bn", "$1tn")
)
test that::expect_equal(
    label_dollar(c(1, 1e3, 1e6, 1e12, 1e18), billion_scale = "long"),
    c("$1", "$1k", "$1m", "$1bn", "$1tn")
)
```

davidchall:
Ping @dpseidel for his opinions, since he originally added the `label_number_si()` function.

With the long/short scale, it seems a distinction can be made between the breaks (long vs short) and the labels themselves (e.g. people use “M”, “m”, “mn”, “million” depending on the setting). If we’re going to rename the current `label_number_si()` function, then perhaps we could also add this functionality?

davidchall:
Pinging @dpseidel for her opinions, since she originally added the `label_number_si()` function.

With the long/short scale, it seems a distinction can be made between the breaks (long vs short) and the labels themselves (e.g. people use “M”, “m”, “mn”, “million” depending on the setting). If we’re going to rename the current `label_number_si()` function, then perhaps we could also add this functionality?

R/label-dollar.R
```diff
@@ -14,6 +14,11 @@
 #'   value is less than `largest_with_cents` which by default is 100,000.
 #' @param prefix,suffix Symbols to display before and after value.
 #' @param negative_parens Display negative using parentheses?
+#' @param billion_scale Named list indicating suffixes given to large values
```
hadley, line 17: Could this be `rescale` or something a bit more generic?

davidchall: Hi @hadley - I renamed the argument to `rescale_large` to be more generic, while still giving some clue as to when it would be used. I also made some other adjustments so it plays nicely with the existing `accuracy` and `scale` arguments.

I think this is ready for another review, when you have time. 👍

R/label-number-si.R
```diff
@@ -17,19 +16,20 @@
 #' demo_continuous(c(1, 1000), label = label_number_si(unit = "m"))
 #'
 #' demo_log10(c(1, 1e9), breaks = log_breaks(10), labels = label_number_si())
+#' demo_log10(c(1e-9, 1), breaks = log_breaks(10), labels = label_number_si(unit = "m"))
 label_number_si <- function(accuracy = 1, unit = NULL, sep = NULL, ...) {
```
davidchall, line 20: Since this PR has already been labeled as a breaking change, would it be ok to also change some default values?

- Could we also elevate `unit` to a required argument? I don't think there is ever a good reason to use `label_number_si()` without a unit, but users can always input an empty string if needed.
- `accuracy = NULL` to automatically choose precision based on data.
``` r
library(scales)

x <- c(1000, 1500, 2000)
label_number_si(accuracy = 1)(x)
#> [1] "1k" "2k" "2k"
label_number_si(accuracy = NULL)(x)
#> [1] "1.00k" "1.50k" "2.00k"
```

<sup>Created on 2021-03-17 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

sjackman: Unitless measure such as counts and ratio seem to me a reasonable use of `label_number_si`. I don't see a strong benefit to making `unit` mandatory.

> `accuracy = NULL` to automatically choose precision based on data.

Could you elaborate? What would the default `accuracy` be?

```
label_number_si(accuracy = NULL)(x)
#> [1] "1.00k" "1.50k" "2.00k"
```

In this example, are you suggesting that it could default to `accuracy=1` rather than the current `accuracy=3`? That seems useful to me.

davidchall: The default value of `accuracy` is currently `1`, but I'm suggesting to change it to `NULL`. This argument is simply passed to the `scales::number()` function. When `accuracy = NULL`, `scales::number()` automatically chooses the precision to display.

The example shows how using `accuracy = NULL` is advantageous. When `accuracy = 1` (the current default), the 2nd and 3rd elements of the output vector are identical.

sjackman: Thank you for the explanation! Yes, `accuracy = NULL` indeed seems like a better default than `accuracy = 1`.

hadley:
It's been a while since I looked at this, so could you please give me a brief summary of the overall goals and what's changed since last time I looked?

davidchall:
Hi @Hadley - sure thing!

## Overall Goals

`label_number_si()` scales each input and assigns a corresponding suffix. For example, `1e3 -> 1 K` (1 thousand) and `1e9 -> 1 B` (1 billion). The problem is that these suffixes do not correspond to the [SI unit prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes) used in science. They actually correspond to the [short scale abbreviations](https://en.wikipedia.org/wiki/Long_and_short_scales) used more colloquially (e.g. finance, populations).

I think it'd be good to retain support for the [short scale](https://en.wikipedia.org/wiki/Long_and_short_scales), but there are some additional complexities to consider. First, many countries use the alternative [long scale](https://en.wikipedia.org/wiki/Long_and_short_scales) or another system (e.g. [Indian numbering system](https://en.wikipedia.org/wiki/Indian_numbering_system)). Because it's particularly common to use the [short scale](https://en.wikipedia.org/wiki/Long_and_short_scales) in finance (e.g. $1M), I decided it makes most sense to add this feature to `label_dollar()`. Other use cases (e.g. populations) could simply disable the dollar prefix.

The goal of this PR is twofold:

1. Adjust `label_number_si()` so it uses the intended SI unit prefixes.
2. Adjust `label_dollar()` so it supports custom scaling-by-suffix. The most common scenarios are supported by `short_scale()` and `long_scale()`.

## Changes since last review

- You requested the `billion_scale` argument of `label_dollar()` be renamed to something more generic, and suggested `rescale`. I've renamed this as `rescale_large`, which hints at when it should be used.
- I adjusted `label_dollar()` so the `accuracy` and `scale` arguments work well with `rescale_large`.
- I changed a default argument of `label_number_si()` to `accuracy = NULL` so that the displayed precision is automatically chosen by `scales::number()`. This is a more natural choice for this labeler following these changes.
- I refactored logic into `rescale_by_suffix()`, so both `label_number_si()` and `label_dollar()` use the same underlying scaling code.
- I requested that we elevate `unit` to a required argument of `label_number_si()`. I don't think it makes sense to use an SI prefix without a unit (though @sjackman thinks this sounds like a reasonable use case). From my perspective, it doesn't make sense to say "5 milli-" without following up with "meter" or "gram" or something else. I was hoping we could help the user to remember to specify the unit. But I don't feel particularly strongly about this.

sjackman:
> a dimensionless quantity is a quantity to which no physical dimension is assigned, also known as a bare, pure, or scalar quantity or a quantity of dimension one,[1] with a corresponding unit of measurement in the SI of the unit one (or 1),[2][3] *which is not explicitly shown* [emphasis added]

https://en.wikipedia.org/wiki/Dimensionless_quantity

It's entirely reasonable to use SI prefixes to represent dimensionless quantities, such as counts and ratios.

sjackman:
Thank you very much for this work, David! It'll make `label_number_si` much more useful to me. Have a good weekend!

sjackman:
I'm glad that you're passionate about SI units! My preference is for `unit` to remain optional, but if the consensus preference is to make it mandatory, then explicitly setting `unit = ""` isn't the worst thing in the world.

Does the official publication make any recommendation as to how to represent dimensionless quantities, other than don't? Which is actually pretty funny. 😆

R/label-dollar.R
```diff
@@ -126,3 +162,15 @@ dollar <- function(x, accuracy = NULL, scale = 1, prefix = "$",
 
   amount
 }
+
+#' @export
+#' @rdname label_dollar
+short_scale <- function() {
```
hadley, line 168: I think these should have a common prefix rather than a common suffix. `scales::scale_short()` seems a bit confusing though. I don't have any better ideas right now, but maybe you do?

R/label-dollar.R
```diff
@@ -102,6 +126,18 @@ dollar <- function(x, accuracy = NULL, scale = 1, prefix = "$",
   negative <- !is.na(x) & x < 0
   x <- abs(x)
 
+  if (!is.null(rescale_large)) {
+    if (!(is.integer(rescale_large) && all(rescale_large > 0))) {
+      stop("`rescale_large` must be positive integers.", call. = FALSE)
+    }
+
+    rescale <- rescale_by_suffix(x * scale, breaks = c(0, 10^rescale_large))
```
hadley, line 134: I think `rescale_by_suffix()` should probably handle the `10^`

R/si-prefixes.R
```diff
@@ -0,0 +1,11 @@
+# power-of-ten prefixes used by the International System of Units (SI)
+# https://www.bipm.org/en/measurement-units/prefixes.html
+#
+# note: irregular prefixes (hecto, deca, deci, centi) are not stored
+# because they don't commonly appear in scientific usage anymore
+si_powers <- (-8:8) * 3
```
hadley, line 6: I think this should live in `label-number-si.R`

hadley (commented):
Thanks for working on this, and for the detailed explanation of what's changed. Overall, it looks great 😀 I have a few minor comments below, but can you please also add a NEWS bullet making it clear that this is a breaking change, and briefly justifying why (i.e. it previously didn't actually use SI units)

R/label-dollar.R
davidchall: Glad you brought this up! It reminded me... 

I meant to ask if I should refactor `label_bytes(units = "auto_si|auto_binary")` to also use `rescale_by_suffix()`? If so, then we don't want `rescale_by_suffix()` to handle the `10^`.

R/label-dollar.R
davidchall: How about if they're called `rescale_short_scale()` and `rescale_long_scale()`? It's a bit verbose, but it aligns with the argument name (`rescale_large`). It also leaves room for things like `rescale_indian()` etc.

R/label-dollar.R
hadley: Yeah, if you think it fits.

R/label-dollar.R
hadley: Yeah, I like it.

davidchall:
@sjackman 

> Does the official publication make any recommendation as to how to represent dimensionless quantities, other than don't? 

There are some types of dimensionless quantities, where the recommendation is to use a derived unit (not a base SI unit). For example, a measurement of plane angle would use radians (which expressed in terms of base SI units is meter / meter). In this situation, we can use the SI prefixes (e.g. milliradians or mrad).

They don't give explicit recommendation for situations where there is no appropriate unit (e.g. count) but the number is significantly different from 1. I think a reasonable approach would be to use [scientific notation](https://en.wikipedia.org/wiki/Scientific_notation). There is already the [scales::label_scientific()](https://scales.r-lib.org/reference/label_scientific.html) formatter for this purpose.

davidchall:
Hi @hadley, I've updated with your feedback. Before updating the NEWS, I was just hoping for your feedback on this point:

 > I requested that we elevate `unit` to a required argument of `label_number_si()`.

The official recommendation from the standards organization is not to use SI prefixes without a unit, so I think 99% of usage will require that this argument is set. My hope is to reduce user error by prompting for this argument. They could still set `unit = ""` if they desire no unit.

@sjackman is in favor of retaining the default value of NULL (i.e. no unit). This has the advantage that existing code still produces a plot (though the result is changed). I think we each have moderate strength of feeling for our own preference, but are both willing to accept the alternative option. So I'm really just looking for a decision on how you'd like {scales} to handle this. There's a bit of discussion above, if you'd like to read more context.

hadley:
Making it required has the nice property that existing users will get an error, forcing them to look up the docs and discover that its changed.

davidchall:
Hi @hadley - I think I'm finished with this PR now. I tried to follow the [NEWS style guide](https://style.tidyverse.org/news.html), but please let me know if I need to update anything. Thanks!

NEWS.md
```diff
@@ -1,8 +1,34 @@
 # scales (development version)
 
+## Breaking changes
```
hadley, line 3: Can you please leave these headings off? I add them just before release. Otherwise, looks good!

davidchall: Ah, sorry for that! I wasn't sure how else to flag them as a breaking change, but it sounds like you'll do this later. Thanks!

sjackman:
Thank you, David! ✨

PR #243, opened 2019-12-18
Handle NA colors correctly in colour_ramp().

### LINKED ISSUE #241: Color interpolation in colour_ramp() can fail in some cases
The colorspace package is [failing testing on CRAN](
https://r-forge.r-project.org/R/?group_id=20&log=check_x86_64_windows&pkg=colorspace&flavor=patched) apparently caused by this line in the code:
https://github.com/r-lib/scales/blob/a33f8ef9d2930f4940f9c694ba6ba79480965cec/R/colour-ramp.R#L58

This is the error:
``` r
> ### ** Examples
> 
> # adapted from stackoverflow: https://stackoverflow.com/a/20127706/4975218
> 
> library("ggplot2")
> 
> # generate dataset and base plot
> set.seed(100)
> df <- data.frame(country = LETTERS, V = runif(26, -40, 40))
> df$country = factor(LETTERS, LETTERS[order(df$V)]) # reorder factors
> gg <- ggplot(df, aes(x = country, y = V, fill = V)) +
+   geom_bar(stat = "identity") +
+   labs(y = "Under/over valuation in %", x = "Country") +
+   coord_flip() + theme_minimal()
>   
> # plot with default diverging scale
> gg + scale_fill_continuous_diverging()
Error in if (!alpha || all(lab_in[, 4] == 1)) { : 
  missing value where TRUE/FALSE needed
Calls: scale_fill_continuous_diverging ...  -> ggproto ->  -> colour_ramp
Execution halted
```

Strangely this error occurs [only on Windows.](https://CRAN.R-project.org/web/checks/check_results_colorspace.html)

It is unclear to me why this happens, since colorspace calls `gradient_n_pal()` which then calls `colour_ramp()` with default arguments here:
https://github.com/r-lib/scales/blob/a33f8ef9d2930f4940f9c694ba6ba79480965cec/R/pal-gradient.r#L16

However, in either case, it looks to me like line 58 in `colour_ramp()` is sensitive to missing `alpha` arguments and would be safer if it read:
``` r
  if (!isTRUE(alpha) || all(lab_in[, 4] == 1)) { 
```

clauswilke:
This closes #241.

``` r
library(scales)

ramp <- colour_ramp(c("red", NA, "blue"))
show_col(ramp(seq(0, 1, length = 12)))
```

<sup>Created on 2019-12-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/colour-ramp.R
+++ b/R/colour-ramp.R
@@ -55,7 +55,9 @@ colour_ramp <- function(colors, na.color = NA, alpha = TRUE) {
   u_interp <- stats::approxfun(x_in, lab_in[, 2])
   v_interp <- stats::approxfun(x_in, lab_in[, 3])
 
-  if (!alpha || all(lab_in[, 4] == 1)) {
+  # need to excludes all NAs inside all(), because all(NA) == NA,
+  # but all(numeric(0)) == TRUE (which is appropriate here)
+  if (!isTRUE(alpha) || all(lab_in[, 4][!is.na(lab_in[, 4])] == 1)) {
     alpha_interp <- function(x) NULL
   } else {
     alpha_interp <- stats::approxfun(x_in, lab_in[, 4])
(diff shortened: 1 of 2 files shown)


hadley:
Superseded by #246

Issue #252, opened 2020-02-04
hue_pal with direction= -1 seems broken


bbarthold:
After switching from `grDevices::hcl` to `farver::encode_colour`, negative hues generated in `hue_pal` function when setting direction to -1 do not seem to work as expected.

See results from running `show_col(hue_pal(direction = -1)(9))` using scales v.1.1.0 (farver 2.0.3) below
``` r
library(scales)
show_col(hue_pal(direction = -1)(9))
```

![](https://i.imgur.com/fZOUc3X.png)

``` r
sessionInfo()
#> R version 3.6.2 (2019-12-12)
#> Platform: x86_64-pc-linux-gnu (64-bit)
#> Running under: Ubuntu 19.10
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/openblas/libblas.so.3
#> LAPACK: /usr/lib/x86_64-linux-gnu/libopenblasp-r0.3.7.so
#> 
#> locale:
#>  [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
#>  [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
#>  [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
#>  [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
#>  [9] LC_ADDRESS=C               LC_TELEPHONE=C            
#> [11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] scales_1.1.0
#> 
#> loaded via a namespace (and not attached):
#>  [1] Rcpp_1.0.3       digest_0.6.23    R6_2.4.1         lifecycle_0.1.0 
#>  [5] magrittr_1.5     evaluate_0.14    highr_0.8        rlang_0.4.4     
#>  [9] stringi_1.4.5    farver_2.0.3     rmarkdown_2.1    tools_3.6.2     
#> [13] stringr_1.4.0    munsell_0.5.0    xfun_0.12        yaml_2.2.1      
#> [17] compiler_3.6.2   colorspace_1.4-1 htmltools_0.4.0  knitr_1.27
```

[...]

PR #253, opened 2020-02-06
Fix direction = -1 functionality for `hue_pal` (#252)

dpseidel:
scales 1.1.0 replaced `grDevices::hcl` with `farver::encode_colour` breaking the `direction  = -1` functionality  for `hue_pal`. Original functionality relied upon parsing negative hue values. 

This new implementation matches our current handling of the direction argument in `brewer_pal` using the `rev` argument to simply reverse the palette returned when `direction = -1`.  This is different from the implementation used in scales 1.0.0 (see below) which returns colors from the opposite side of the color wheel, but I think this actually is more consistent with expected `direction = -1` behavior for discrete scales. 

#### `hue_pal(direction = 1)`

``` r
show_col(hue_pal(direction = 1)(4))
```

#### `hue_pal(direction = -1)`

scales 1.0.0 

``` r
show_col(hue_pal(direction = -1)(4))
```

scales 1.1.0

``` r
show_col(hue_pal(direction = -1)(4))
```

scales 1.1.0.9000 (dpseidel/scales, patch_huepal)

image

This PR addresses #252

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # scales (development version)
 
+* `hue_pal` now correctly inverts color palettes when `direction = -1` (@dpseidel, #252). 
+
 # scales 1.1.0
 
 * Axis breaks and labels have a new naming scheme: functions that generate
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!

Issue #255, opened 2020-02-15
Feature request: Inert out-of-bounds function


teunbrand:
I think it might be useful to have an out-of-bounds function similar to `censor()`, `squish()` and `squish_infite()` that does nothing to the data. The function can be very simple and trivial (probably could use a better name though):

```r
inert_oob <- function(x, ...) x
```

It is very similar to `base::identity` and `base::force` but these don't take additional arguments and raise errors when used as `oob`-arguments in ggplot2. I'm sure a function like this must exist somewhere, but maybe my R vocabulary isn't large enough.

The use case for this in the ggplot2 package, would be to achieve the same what the coordinate limits do to the plot.

To demonstrate this, notice that the following plots are equivalent:

```r
df <- data.frame(
  x = 1:3,
  y = c(1, 10, Inf)
)

ggplot(df, aes(x, y)) +
  geom_col() +
  coord_cartesian(ylim = c(0, 5))
```
![image](https://user-images.githubusercontent.com/49372158/74592060-dc637280-501d-11ea-8799-08e943411daf.png)

```
ggplot(df, aes(x, y)) +
  geom_col() +
  scale_y_continuous(
    limits = c(0, 5),
    oob = inert_oob
  )
```
![image](https://user-images.githubusercontent.com/49372158/74592074-f604ba00-501d-11ea-9879-8bd4a99964cc.png)

I've seen on quite a number of stack overflow posts that people are confused why they must use the coordinate limits and not the scale limits or can't figure out why their barplots dissappear when the lower-limit on y is > 0 (for example: https://stackoverflow.com/questions/58954151/problems-adapting-the-y-axis-to-2x2-anova-bargraph-using-r-and-ggplot/)

Formalising this behaviour of the scale limits as a function might help people better understand that not touching the data is a valid option for axis limits too.

If this function already exists somewhere, maybe it might be best to ignore this issue and I should ask the people at ggplot2 whether they would consider to document this use somewhere.

Thank you for taking the time to read!

clauswilke:
This sounds like a good idea to me. Just needs a better name. `leave_alone()`? `keep_all()`?

teunbrand:
Those both seem better than `inert_oob()`. The censor/squish/squish_infinite seems to be verbs that describe what is being done to the out of bounds values, i.e.: censor out of bounds values, squish out of bounds values. 

[...]

PR #258, opened 2020-02-21
Adds keep_all as bounds function (#255)

teunbrand:
For the request in #255; adds an out of bounds function that keeps values outside the range, i.e. does nothing. 
Using `keep_all` as `oob` argument in e.g. `scale_x_continuous()` would in ggplot2 be the equivalent to setting the coord limits.


teunbrand:
Fixes #255

PR #266, opened 2020-04-03
add optional ncol argument to show_col.

### LINKED ISSUE #250: Expose n_col in show_col
Today in `show_col()`
```r
show_col <- function(colours, labels = TRUE, borders = NULL, cex_label = 1) {
  n <- length(colours)
  ncol <- ceiling(sqrt(n))
  nrow <- ceiling(n / ncol)
...
```
I am sometimes printing multiple color palettes and would like to control how they are displayed. Could we please expose optional `ncol` argument with default value as it is today.
```r
show_col <- function(colours, labels = TRUE, borders = NULL, 
                    cex_label = 1, ncol = ceiling(sqrt(length(colours)))) 
```
Alternative, but perhaps equally useful feature could be to allow passing a character matrix (vector with `dim` attribute) and let that control the shape of the printed palette. (I don't know if you considered showing labels, if a vector is named)

I am also a little concerned about the margins of the plot when number of colors is small. It seems like we could make margins tighter by calculating anticipated size of the plot. 

Let me know if you would like a PR with this.

dmi3kno:
Closes #250.

### DIFF
--- a/R/colour-manip.r
+++ b/R/colour-manip.r
@@ -74,17 +74,26 @@ alpha <- function(colour, alpha = NA) {
 #'
 #' A quick and dirty way to show colours in a plot.
 #'
-#' @param colours a character vector of colours
-#' @param labels boolean, whether to show the hexadecimal representation of the colours in each tile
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

hadley:
Thanks!

dmi3kno:
Thanks for improving the function so much!

Issue #267, opened 2020-04-05
Document all oob_ functions in one file


hadley:
Since they're all quite simple

clauswilke:
@teunbrand Are you willing to take this on?

teunbrand:
Sure, seems a good idea to me. I'll write something up!

teunbrand:
A thought occurred to me that is tangentially related to this issue. Currently there are two versions of squishing values inside the range: `oob_squish()` and `oob_squish_infinite()`. 

We could consider splitting out `oob_squish` to `oob_squish_finite()`, with the current default behaviour, and `oob_squish_all()`, which would behave as `oob_squish(x, only.finite = FALSE)`. 

I may assume wrongly here, but my reasoning is that if the main user interaction with these functions is to pass these as `oob` arguments to continuous/binned scales, then the `only.finite` argument to `oob_squish()` is impractical to set. You cannot —or I wouldn't immediately know how to— pass this function with that argument without a small wrapper; for example `function(x, range){oob_squish(x, range, only.finite = FALSE)}`. Then, it would be more convenient to have a dedicated function for this.

I could post this as a seperate issue and weave this into the PR if you think that makes sense.

clauswilke:
This makes sense to me. Would you keep `squish` as an alias to `oob_squish_finite` for backwards compatibility?

hadley:
Fixed in #268

PR #268, opened 2020-04-05
Shared documentation for oob_functions. Split oob_squish.

teunbrand:
As suggested in #267, documented all out of bounds (`oob_*()`) functions in one page. Descriptions of individual functions are included through the `@describeIn` tag. Also gathered relevant unit tests and included a few new expectations for `oob_squish_infinite()`, `oob_discard()` and old interfaces.

Also, I split `oob_squish` to `oob_squish_any()` and `oob_squish_finite()`, where the latter is backwards compatible with previous `squish()` usage. Would have liked to remove the `only.finite` argument of `oob_squish_finite()`, but didn't seem like the wisest choice for backward compatibility.


clauswilke:
Looking at this now, I'm wondering whether we're handling the squish situation correctly. I notice that `oob_censor()` has the same `only.finite` argument. So that would suggest adding `oob_censor_finite()` and `oob_censor_any()` but that seems overly cumbersome. If we feel the "finite" version is the default, why not keep `oob_squish()` and `oob_censor()` and then just add `oob_squish_any()` and `oob_censor_any()` as shortcuts for the case with `only.finite = FALSE`?

This also helps with the issue that a naive user might think `oob_squish_infinite()` is the opposite of `oob_squish_finite()` but if I read the code correctly that's not the case.

Finally, why not define `oob_squish_any()` as simply a wrapper around `oob_squish()` setting `only.finite = FALSE`? I realize that's a tiny bit slower but feels safer to me implementation-wise.

R/bounds.r
```diff
@@ -155,17 +155,52 @@ rescale_none <- function(x, ...) {
...
+#' @param x A numeric vector of values to modify.
+#' @param range A numeric vector of length two giving the minimum and maximum of
+#'   the desired output range respectively.
+#' @param only.finite A logical of length one, which when `TRUE` will only alter
+#'   finite values. When `FALSE`, will also alter infinite values.
+#'
+#' @return Most `oob_()` functions return a vector of numerical values parallel
+#'   to the '`x`' argument, wherein out of bounds values have been modified.
+#'   Only `oob_discard()` returns a vector of less than or of equal length to
+#'   the '`x`' argument.
```
clauswilke, line 173: Not sure why we need the single quotes around `x` here (two occurrences).

teunbrand:
You make some good points, I'll adjust it.

R/bounds.r
```diff
@@ -155,17 +155,52 @@ rescale_none <- function(x, ...) {
...
+#' @name oob
+#' @aliases oob_censor censor oob_squish_finite oob_squish_any squish
+#'   oob_squish_infinite squish_infinite oob_discard discard oob_keep
+#' @title Out of bounds handling
+#' @description This set of functions modify data values outside a given range.
+#'
+#' @param x A numeric vector of values to modify.
+#' @param range A numeric vector of length two giving the minimum and maximum of
+#'   the desired output range respectively.
+#' @param only.finite A logical of length one, which when `TRUE` will only alter
```
clauswilke, line 167: I realize this is just copied over from the previous version, but I don't like the phrasing here. As written, the sentence implies that the parameter `only.finite` alters the data, which is not true. It's the function that alters the data. A simple way to fix this may be to use passive voice: "A logical of length one. When set to `TRUE` only ... will be altered ..."

teunbrand:
I think I've adressed most of your comments in the latest commit. I don't know why the R >3.2 ubuntu checks fail though.

clauswilke:
The ubuntu tests fail at the installation stage because a required repo is not available. Has nothing to do with your PR specifically, as far as I can tell.

R/bounds.r
```diff
@@ -155,56 +155,91 @@ rescale_none <- function(x, ...) {
   x
 }
 
-#' Censor any values outside of range
+#' @name oob
+#' @aliases oob_censor oob_censor_any censor oob_squish oob_squish_any squish
+#'   oob_squish_infinite squish_infinite oob_discard discard oob_keep
+#' @title Out of bounds handling
```
hadley, line 161: Our general style is not to use `@title` — it should just be the first line of the docs

R/bounds.r
```diff
@@ -155,56 +155,91 @@ rescale_none <- function(x, ...) {
...
+#'   the desired output range respectively.
+#' @param only.finite A logical of length one. When `TRUE`, only finite values
+#'   are altered. When `FALSE`, also infinite values are altered.
+#'
+#' @return Most `oob_()` functions return a vector of numerical values parallel
+#'   to the `x` argument, wherein out of bounds values have been modified. Only
+#'   `oob_discard()` returns a vector of less than or of equal length to the `x`
+#'   argument.
+#'
+#' @details The `oob_*()` functions are designed to be passed as `oob` argument
```
hadley, line 175: I'd suggest combining this with `@description` (so it's "above the fold" in the docs) and use a bulleted list to make it easier to scan the functions

R/bounds.r
```diff
@@ -155,56 +155,91 @@ rescale_none <- function(x, ...) {
...
+#' @title Out of bounds handling
+#' @description This set of functions modify data values outside a given range.
+#'
+#' @param x A numeric vector of values to modify.
+#' @param range A numeric vector of length two giving the minimum and maximum of
+#'   the desired output range respectively.
+#' @param only.finite A logical of length one. When `TRUE`, only finite values
+#'   are altered. When `FALSE`, also infinite values are altered.
+#'
+#' @return Most `oob_()` functions return a vector of numerical values parallel
```
hadley, line 170: parallel to -> the same length as

R/bounds.r
```diff
@@ -155,56 +155,91 @@ rescale_none <- function(x, ...) {
...
-  # Assign NA - this makes sure that, even if all elements are
-  # replaced with NA, it stays numeric (and isn't changed to logical)
   x[finite & x < range[1]] <- NA_real_
   x[finite & x > range[2]] <- NA_real_
   x
 }
 
-#' @rdname oob_censor
+#' @export
+#' @describeIn oob replaces out of bounds values and infinite values with
```
hadley, line 219: I think you can omit these `@describeIn` in favour of one bulleted list.

teunbrand:
Thanks Claus and Hadley for the comments. I've adressed them in the latest commit.

clauswilke:
It looks to me like the latest version still doesn't quite follow the style guide. We never use the `@title` tag. See e.g. this example:

https://github.com/r-lib/scales/blob/472c466ac96f5b37283f3dac6d3f2edd194f2b73/R/colour-ramp.R#L1-L9

teunbrand:
True, I accidentally let these tags slip through the cracks; my apologies.

hadley:
Thanks @teunbrand!

PR #269, opened 2020-04-06
Support units for `breaks_width(offset = )` #247

nacnudus:
Based on `fullseq()` in  `R/full-seq.r`.

`base::seq()` is used for consistency with the `width` argument.  To add one
unit, a sequence of `length.out = 2, by = unit` is created, and the second
element is taken.  This process is vectorised using `lapply()` and `do.call(c)`,
because the alternatives `vapply()` or `unlist()` strip the attributes of
`Date` and `POSIXt` objects.

This is modified to apply `offset_by()` to the output of `fullseq()`.  To
support offstes with compound units, for example one month and five days for the
UK tax year, the `offset` argument can be a vector, and the offsets are applied
sequentially in the order given.  An example illustrates the difference that the
order can make.


hadley:
@nacnudus basic approach looks good. Could you please add a couple of tests, primarily to `test-offset-by.R`?

Note that I made a few changes (fixing the whitespace mangling in NEWS, and simplifying the docs a bit), so you'll need to `usethis::pr_pull()` (or similar) to get my changes locally.

hadley:
Also, could you please add an `difftime` method to `offset_by()` since we now support difftime objects?

nacnudus:
* Added tests of `offset_by()`
* Added `offset_by.difftime()`
* Simplified `offset_by.numeric()`
* Removed dots from `offset_by()` args. I think I'd blindly copied them from somewhere else.

hadley:
Thanks @nacnudus!

Issue #249, opened 2020-01-30
Support Unicode-2212 Minuses


BenjaminWolfe:
Huge fan of the scales package.

In the last year or two I learned the hyphen minus (ASCII 45) and true minus (Unicode 2212) are different characters. The true minus is basically the plus without the vertical bar: It has the same length, thickness, and positioning as the plus's crossbar. So it looks really nice / professional with numbers and plots. It's also friendly to screen readers. But when I use the scales package as is I get hyphen-minuses.

I made the [signs package](https://github.com/BenjaminWolfe/signs) to address this for anyone who's as much of an unabashed typography nerd as I am. But I only went with of a separate package because a) I was totally new to open-source contribution, and it was somewhat scary to touch someone else's work, especially work of which I'm such a fan; b) in the meantime I needed to use the functionality in my own work; and quite honestly c) I really thought only a few people would find the case compelling!

But the better way to do this, and one that will make me feel like a *real* citizen of the open-source community, would be to (also) submit a pull request to add the functionality to the scales package. And at rstudio::conf I met people who expressed interest in the difference between the characters. So I would truly love to work on this PR. (I'll add that I don't *have* to be the one working on it, if someone else on the team would prefer to take it up.)

Before I jump in, I'm guessing the right path is to open this issue and allow some discussion on how to proceed? I imagine I can save some headaches this way. My questions are...

- Any user-oriented design considerations, i.e. how you want the user to interact with it? I imagine making it as transparent as possible, maybe implementing it as an option users can set at the top of a script.
- What work would be involved? I.e. for someone who knows scales better than I, what's the quick outline of chunks of work I should do?
- And then: How can I truly best contribute to the work? Is it as simple as forking, trying it out, making sure all tests pass, and make a PR? Anything I should know?
- And, well... Anyone I might talk to mentor me through the process a bit? Any time I have this many questions on something, it's a clue that it might be helpful to just have a person to chat with. (Maybe @clauswilke... seemed interested in the use case, and I just learned we both live in Austin!!)

(Someday if I'm a more practiced open-source contributor than I am today, I expect this issue will remind me *just* how many unknowns there are for a neophyte!)

clauswilke:
I think the right approach is to add this feature to the `number()` function:
https://github.com/r-lib/scales/blob/be540d3a2babf145b6b30e0ac1ab87f0f460e404/R/label-number.r#L145-L171
There should probably be an option that allows me to turn this on or off, and/or to specify the character to use as a minus sign.

[...]

PR #280, opened 2020-06-18
Allow functions for prefix and suffix; closes #262

### LINKED ISSUE #262: label_number() argument to always print number with sign
Currently, `label_number()` only prints negative numbers with a sign. I'm requesting an argument to force a `+` symbol before positive numbers.

```r
> scales::label_number()(seq(-1, 1, 0.5))
[1] "-1.00" "-0.50" "0.00"  "0.50"  "1.00"
```

The base R `format()` method doesn't provide an argument for this, so it's not currently possible to achieve this with dot args. Probably it would need to be added in post-processing between these lines (i.e. before adding a prefix):

https://github.com/r-lib/scales/blob/1e2c9186efce35b59a30d3b98e9e50ad0d555512/R/label-number.r#L154-L163

One open question is whether zero should get `+`, `-` or nothing? The `sprintf()` function suggests that we should use `+`. This is also consistent with Python and C.

```r
> sprintf("%+3.2f", seq(-1, 1, 0.5))
[1] "-1.00" "-0.50" "+0.00" "+0.50" "+1.00"
```

dkahle:
I don't really know what the use is for allowing prefix and suffix to be functions is other than it will allow you to add `+`'s before positive percents. I've added `plus = TRUE` to percent functions `percent()` and `label_percent()` for easy use. Happy to adjust!, and sorry about re-roxing things.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 # scales 1.1.1
 
+* `number()` and reliant label functions now allow functions for the prefix and 
+  suffix arguments (@dkahle, #262).
+
 * `breaks_width()` now handles `difftime`/`hms` objects (@bhogan-mitre, #244).
 
 * `hue_pal()` now correctly inverts color palettes when `direction = -1` 
--- a/R/label-number.r
+++ b/R/label-number.r
@@ -4,18 +4,16 @@
 #' [scientific][label_scientific] notation). `label_comma()` is a special case
 #' that inserts a comma every three digits.
 #'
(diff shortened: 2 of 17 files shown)


dkahle:
Another thought is that you may want to allow __rlang__ style anonymous functions, but that would require importing __rlang__. I didn't implement that part, but if you're interested in that, let me know.

hadley:
Closing because while this resolves #262, we now need to also consider #249.

PR #283, opened 2020-06-25
Add na.rm support to DiscreteRange$train

dpseidel:
DiscreteRange train function does not currently support the `na.rm` argument to `train_discrete`,  preventing its direct use to replace ggplot2's range functions (tidyverse/ggplot2#2710). This PR updates the train function to support this argument.

### DIFF
--- a/R/range.r
+++ b/R/range.r
@@ -17,8 +17,8 @@ DiscreteRange <- R6::R6Class(
   "DiscreteRange",
   inherit = Range,
   list(
-    train = function(x, drop = FALSE) {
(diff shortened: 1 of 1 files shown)


dpseidel:
@clauswilke are you concerned by the failing check? looks to be related to a warning regarding links on the brewer_pal documentation. Not entirely sure what's going on there but it's definitely unrelated to this PR. If not, I'll merge.

clauswilke:
Cross-references in documentation are an ongoing drama, but that shouldn't hold up this PR. Not sure why it hits this particular test, but maybe just because {plyr} isn't currently available for R devel.

PR #291, opened 2020-07-23
Unname the result of manual_pal()

### LINKED ISSUE #284: Unname the result of manual_pal()
(Originally discussed at https://github.com/tidyverse/ggplot2/issues/4087#issuecomment-649204121)

ggplot2's `scale_manual()` family uses two types of mapping colours to values; by position and by name. `manual_pal()` is meant to be used for the latter one. But, since the palette `manual_pal()` provides aren't unnamed, it might be accidentally mapped in by-name manner (See tidyverse/ggplot2#4087 for an example). So, I think it should be `unname()`ed as the names are not useful anyway, I believe.

If this looks good, I'll create a pull request for this.

yutannihilation:
Close #284

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # scales (development version)
 
+* `manual_pal()` now always returns an unnamed colour vector, which is easy to
+  use with `ggplot2::discrete_scale()` (@yutannihilation, #284).
+
 # scales 1.1.1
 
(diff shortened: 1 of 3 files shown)


yutannihilation:
Sorry, I forgot to add it. Here it is.

hadley:
Thanks!

PR #292, opened 2020-08-27
hue_pal() now uses argument h.start

### LINKED ISSUE #288: scales v1.1.1.: pal_hue() does not use argument 'h.start' anymore
The pal_hue() function does not use the argument 'h.start' anymore.
Compare:

https://github.com/r-lib/scales/blob/v1.1.1/R/pal-hue.r

with

https://github.com/r-lib/scales/blob/v1.1.0/R/pal-hue.r

'h.start' is not applied anymore in the latest version (v1.1.1). 

If this was by intention, my guess is that 'h.start' should also be removed from the arguments lists, since it does not do anything.
Otherwise, this probably shows the need of introducing a test showing that the argument is indeed used.

Kind regards,
Angela

aalucaci:
closes #288

### DIFF
--- a/R/pal-hue.r
+++ b/R/pal-hue.r
@@ -36,7 +36,9 @@ hue_pal <- function(h = c(0, 360) + 15, c = 100, l = 65, h.start = 0, direction 
       h[2] <- h[2] - 360 / n
     }
 
-    hues <- seq(h[1], h[2], length.out = n)
+    do_shift <- function(x) (x + h.start)
+    hues <- do_shift(seq(h[1], h[2], length.out = n))
+
     hcl <- cbind(hues, c, l)
 
     pal <- farver::encode_colour(hcl, from = "hcl")
--- a/tests/testthat/test-pal-hue.r
+++ b/tests/testthat/test-pal-hue.r
@@ -20,3 +20,17 @@ test_that("hue_pal respects direction argument #252", {
   expect_equal(col1(3), rev(col2(3)))
   expect_equal(col1(9), rev(col2(9)))
 })
+
+
+test_that("hue_pal respect h.start argument #288", {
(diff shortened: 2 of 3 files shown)


tabassco:
This seems to fix the problem described in #288. If someone of the maintainers could shed a glance, it would be greatly apreciated.

R/pal-hue.r
```diff
@@ -36,7 +36,9 @@ hue_pal <- function(h = c(0, 360) + 15, c = 100, l = 65, h.start = 0, direction
       h[2] <- h[2] - 360 / n
     }
 
-    hues <- seq(h[1], h[2], length.out = n)
+    do_shift <- function(x) (x + h.start)
+    hues <- do_shift(seq(h[1], h[2], length.out = n))
```
hadley, line 40: Why not do `seq(h[1], h[2], length.out = n) + h.start`? And does this need rescaling to make sure that `hues` doesn't end up `> 360`?

tests/testthat/test-pal-hue.r
```diff
@@ -20,3 +20,17 @@ test_that("hue_pal respects direction argument #252", {
...
+
+test_that("hue_pal respect h.start argument #288", {
+  col1 <- hue_pal(
+    h = c(0, 360) + 15, c = 100, l = 65, h.start = 0, direction = 1
+    )(9)
+  col2 <- hue_pal(
+    h = c(0, 360) + 15, c = 100, l = 65, h.start = 15, direction = 1
+  )(9)
+
+  expect_true(
```
hadley, line 33: It would be much better to be precise here, as there are many many ways for two things to not be equal.

hadley:
Thanks for working on this, but since I was in the guts of scales, I figured it'd just be easiest to fix myself.

PR #300, opened 2020-11-14
label_bytes() fixes

### LINKED ISSUE #299: label_bytes() symbol for kibibytes
Currently `label_bytes()` uses the symbol `kiB` for kibibytes. Shouldn't this be `KiB` instead?

mikmart:
Some fixes to `label_bytes()`:

1.  Use "KiB" for kibibytes, rather than "kiB". Fixes #299.
2.  Use "B" for bytes when given `units = "auto_binary"`, rather than "iB".
3.  Allow passing `units = "B"`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -38,7 +38,7 @@
 
 * New `label_bytes()` replaces `number_bytes_format()` with a more 
   convenient interface. It takes a single `unit` argument which can either be
-  an SI unit (e.g. "kB"), a binary unit (e.g. "kIB"), or an automatic unit
+  an SI unit (e.g. "kB"), a binary unit (e.g. "KiB"), or an automatic unit
   (either "auto_si" or "auto_binary").
   
     It always uses "B" as the symbol for bytes (#174), and checks that `units` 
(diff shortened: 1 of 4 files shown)


rodarima:
Why is this not merged? I cannot see the failing log test. I would like to fix the "iB" prefix showing on my plots, and the Kibibyte symbol as defined by ISO/IEC 80000-13 should be "KiB" not "kiB" (but kB is correct when using the SI).

mikmart:
If I remember correctly, the one failing R CMD check was due to a failed dependency installation on the CI server, so not related to the changes. That would probably be fixed by just re-running the tests.

rodarima:
Hmm, I read [here](https://stackoverflow.com/a/55419802) that you can just close and reopen the PR to trigger a rerun with the same commit.

mikmart:
Didn't seem to work -- I guess it's dependent on configuration.

mikmart:
An empty commit did the trick. However, looks like there's been changes that have now introduced a merge conflict.

I don't think I'll get around to resolving that any time soon, so I'm going to close this PR. @rodarima feel free to use this branch as a starting point if you want to open a new PR about these issues.

PR #304, opened 2021-03-27
Adjust auto accuracy heuristic

### LINKED ISSUE #279: number(accuracy = NULL) heuristic too precise?
``` r
scales::number(c(10, 20, 30))
#> [1] "10" "20" "30"
scales::number(c(5, 10, 15))
#> [1] "5.0"  "10.0" "15.0"
```

When the smallest interval is >= 1 and < 10, the heuristic computes an accuracy of 0.1.

This might be the intended behaviour, but I find it puzzling: why would the extra decimal be desirable?

The question [has been raised in passing before](https://github.com/r-lib/scales/issues/251#issuecomment-586699978), but I didn't find any other discussion.

If this behaviour is modified, it might also make sense to change the default in `label_number_si()` to `accuracy = NULL`, since the current default of `accuracy = 1` gives results like:

``` r
scales::label_number_si()(c(1.5E6, 2E6, 2.5E6))
#> [1] "2M" "2M" "2M"
```

davidchall:
Fixes #279

The auto accuracy heuristic shows 2 significant figures, relative to the smallest difference between breaks. This is designed to handle situations like (taken from [tidyverse blog post](https://www.tidyverse.org/blog/2019/11/scales-1-1-0/)):

However, this heuristic displays an unnecessary digit in the specific case when all the breaks align with zeros in the final digit. This situation naturally occurs very often. Here's an example (taken from the [scales docs](https://scales.r-lib.org/reference/label_number_si.html)):

image

This PR reduces the precision when final digit is always zero.

## Before

``` r
library(scales)

number(seq(0, 0.1, 0.025))
#> [1] "0.000" "0.025" "0.050" "0.075" "0.100"
number(seq(0, 20, 5))
#> [1] "0.0"  "5.0"  "10.0" "15.0" "20.0"
```

<sup>Created on 2021-03-27 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

## After

``` r
library(scales)

number(seq(0, 0.1, 0.025))
#> [1] "0.000" "0.025" "0.050" "0.075" "0.100"
number(seq(0, 20, 5))
#> [1] "0"  "5"  "10" "15" "20"
```

<sup>Created on 2021-03-27 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,9 @@
   
 * `label_bytes()` now correctly accounts for the `scale` argument when choosing
   auto units (@davidchall, #235).
+  
+* Internal `precision()`, used when `accuracy = NULL`, now avoids displaying
+  unnecessary digits (@davidchall, #304).
 
 # scales 1.1.1
 
--- a/R/label-number.r
+++ b/R/label-number.r
@@ -186,7 +186,15 @@ precision <- function(x) {
   if (smallest_diff < sqrt(.Machine$double.eps)) {
     1
   } else {
+    precision <- 10^(floor(log10(smallest_diff)) - 1)
+
+    # reduce precision when 2nd sig fig always 0
+    reduce_precision <- all(round(x / precision) %% 10 == 0)
+    if (reduce_precision) {
+      precision <- precision * 10
+    }
+
     # Never return precision bigger than 1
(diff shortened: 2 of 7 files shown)


hadley:
Thanks! This is a great little improvement.

Issue #311, opened 2021-06-15
trans_format and label_math behaving differently on log10 scale 


fontikar:
---
author: fontikar
date: 2021-06-15
output: "reprex::reprex_document"
title: drear-kitty_reprex.R
---
First time issue lodger so forgive me if I am not thorough enough. Feedback welcome!
I've been trying to convert retired scales functions to use the 'new' recommended functions for a particular function in a R package I am developing (https://github.com/traitecoevo/austraits/blob/develop/R/plot_trait_distribution_beeswarm.R). 
I noticed label_math() is not behaving the same as trans_format() - code and plots to illustrate below. 
Any advice would be greatly appreciated!

``` r
library(scales)
library(patchwork)

# Old retired code that needs to be updated following scales recommendations
#"parse_format() and math_format() was retired; please use label_parse() and label_math() instead"
demo_log10(x = c(0.1, 10e07),
           breaks = scales::trans_breaks("log10", function(x) 10^x),
           labels = scales::trans_format("log10", scales::math_format(10^.x)))
#> scale_x_log10(breaks = scales::trans_breaks("log10", function(x) 10^x), 
#>     labels = scales::trans_format("log10", scales::math_format(10^.x)))
```

![](https://i.imgur.com/MSV5znn.png)

``` r
#Lets do the breaks first
demo_log10(x = c(0.1, 10e07),
           breaks = trans_breaks("log10", function(x) 10^x)) #Old retired code
#> scale_x_log10(breaks = trans_breaks("log10", function(x) 10^x))
```

![](https://i.imgur.com/WQjvtKq.png)

``` r
demo_log10(x = c(0.1, 10e07),
           breaks = breaks_log(6)) #YAY easy conversion
#> scale_x_log10(breaks = breaks_log(6))
```

![](https://i.imgur.com/eLBkvof.png)

``` r
#Now the labels
demo_log10(x = c(0.1, 10e07),
           breaks = breaks_log(6), 
           labels = trans_format("log10", scales::math_format(10^.x))) #Old retired code
#> scale_x_log10(breaks = breaks_log(6), labels = trans_format("log10", 
#>     scales::math_format(10^.x)))
```

![](https://i.imgur.com/OAOvxr8.png)

``` r
lab_math <- demo_log10(x = c(0.1, 10e07),
           breaks = breaks_log(6), 
           labels = scales::label_math()) #The superscript is not what is expected
#> scale_x_log10(breaks = breaks_log(6), labels = scales::label_math())

[...]

PR #312, opened 2021-06-24
Add `label_log()`

davidchall:
@clauswilke - as discussed in #311. Please let me know what you think.

The default behavior of `label_log()` is essentially an alias for `label_math(expr = 10^.x, format = log10)`. But it also has a `base` argument to support other logarithm bases (see examples below).

~Since the next release already contains breaking changes, I took the liberty of removing the default expression from `label_math()`. I don't see any reason for this default expression over any others. But please let me know if this change should be reverted.~

``` r
library(scales)

demo_log10(c(1, 1e5), labels = label_log())
#> scale_x_log10(labels = label_log())
```

``` r
demo_log10(c(1, 1e5), breaks = breaks_log(base = 2), labels = label_log(base = 2))
#> scale_x_log10(breaks = breaks_log(base = 2), labels = label_log(base = 2))
```

<sup>Created on 2021-06-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


clauswilke:
>  I took the liberty of removing the default expression from `label_math()`

That's a question for @hadley, but I'd probably keep the current default it since tons of code may depend on it.

davidchall:
> >  I took the liberty of removing the default expression from `label_math()`
> 
> 
> 
> That's a question for @hadley, but I'd probably keep the current default it since tons of code may depend on it.

Ah, I see this function has been around for a long time. Ok - I can reverse that change.

caparks2:
Can the concept of `label_log()` be extended to include formatting for other log transformations? It would be particularly helpful if it were to also work with `pseudo_log_trans()`. Use of `label_math(expr = 10^.x, format = log10)` plays havoc with the exponent values after using the pseudo log transform in the current stable release of scales.

R/label-log.R
```diff
@@ -0,0 +1,25 @@
...
+#'
+#' @param base base of logarithm to use
+#' @inherit number_format return
+#' @seealso [breaks_log()] for the related breaks algorithm.
+#' @export
+#' @family labels for log scales
+#' @examples
+#' demo_log10(c(1, 1e5), labels = label_log())
+#' demo_log10(c(1, 1e5), breaks = breaks_log(base = 2), labels = label_log(base = 2))
+label_log <- function(base = 10) {
```
hadley, line 13: I think it would be good to add a `digits` argument here and then explicitly call `format(log(x, base = base), digits = digits)` to make it more clear where the conversion from number to text is happening.

PR #323, opened 2021-12-28
Fix `label_number_auto()` to work on single values > 1e+06 

### LINKED ISSUE #321: label_number_auto() fails on certain specific values
Under scales 1.1.1, `label_number_auto()` fails on certain specific numbers.

``` r
dat <- c(1636500, 30925005)

#works
scales::label_number_auto()(dat)
#> [1] "1 636 500"  "30 925 005"

#fails
scales::label_number_auto()(dat[2])
#> Error in apply(labels, 1, shortest): dim(X) must have a positive length

#fails
scales::label_number_auto()(30925005)
#> Error in apply(labels, 1, shortest): dim(X) must have a positive length
```

I'm specifically trying to apply `label_number_auto()` to a one row tibble with some of these problematic values in it via `across()`.

Debugging this, it looks like in the `format_shortest` call:

https://github.com/r-lib/scales/blob/9c5a00d66d3536e17f130abf50af9c3a61e1059d/R/label-number-auto.R#L53-L57

...the `labels` object is getting passed as a vector, which `apply` fails on. Some testing with ```   if (is.null(dim(labels))) {
      labels <- matrix(labels, nrow = 1)
    }``` looks like a possible work around, but the threads on how apply deals with vectors is a bit of a mystery to me.

Please let me know if I can add anything to help clarify this issue. Thanks!

karawoo:
Fixes #321. `vapply()` would return a character vector instead of a matrix when formatting a single value, which then caused `apply()` to throw an error. This ensures that we're calling `apply()` on a matrix with the necessary dimensions.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # scales (development version)
 
+* `label_number_auto()` correctly formats single numbers that are greater than
+  1e+06 without an error (@karawoo, #321)
+
 * `manual_pal()` now always returns an unnamed colour vector, which is easy to
   use with `ggplot2::discrete_scale()` (@yutannihilation, #284).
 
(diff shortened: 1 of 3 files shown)


hadley:
Thanks @karawoo!

PR #330, opened 2022-03-17
Add locale argument to date/time formatters

### LINKED ISSUE #309: Ability to change date language on axis
Currently it's not possible to change the date language for an axis. If a graph needs to have months in text (ie "February 1", not "2/1"), there's no way to ask ggplot to write in non-English. It is possible via a VERY complex hardcode using glue and a manually written spreadsheet, the result is the first graph in this blog post: https://www.hiringlab.org/fr/blog/2020/11/06/coronavirus-et-offres-demploi-au-30-octobre/. 

Having the ability to set the language so the months on an axis appear in a non-English language would be amazing. The ability to specific American vs European date format is also important. February 1 is fine for North American graphs but for Europe is should return 1 février (in the French case).

Our team would love to have:
French, German, Spanish, Italian, Polish, Portuguese, Dutch, Flemish

Of course any other languages would be great too!!

hadley:
Fixes #309

@DavisVaughan do you think it's ok to add a soft stringi dependency here? It seems like the least worst solution to me as if we rely on `withr::local_locale()` there's no easy way to get a listing of your options and you have to use different locales on windows. I also looked at `readr::date_names()`, but there's no obvious way to use that for formatting.


R/label-date.R
```diff
@@ -113,6 +121,15 @@ label_time <- function(format = "%H:%M:%S", tz = "UTC") {
   }
 }
 
+format_dt <- function(x, format, tz = "UTC", locale = NULL) {
+  if (is.null(locale)) {
+    format(x, format = format, tz = tz)
+  } else {
+    check_installed("stringi")
+    format <- stringi::stri_datetime_fstr(format)
+    stringi::stri_datetime_format(x, format, tz = tz, locale = locale)
```
DavisVaughan, line 130: I think the main thing to be aware of is the fact that `stri_datetime_format()` doesn't support the exact same format specifiers as `format.POSIXct()`. i.e. in `?stri_datetime_format`:

> Warnings are given in the case of %x, %X, %u, %w, %g, %G, %c, %U, and %W

It also won't support fractional seconds like `%OS5`, that's special to R and is fairly non-standard.

It also doesn't support these super special locale specific ones, i.e. in `?strptime`:

> For output there are also %O[dHImMUVwWy] which may emit numbers in an alternative locale-dependent format (e.g., roman numerals), and %E[cCyYxX] which can use an alternative ‘era’ (e.g., a different religious calendar). Which of these are supported is OS-dependent. 

---

You could also use clock if you wanted to. It has `clock_locale()`, which is like `readr::date_names()`, and you use the locale object with `date_format()`. This also means users could provide a custom locale object with their own month and weekday names if they wanted to, but you'd have to document that you use `clock::clock_locale()` to create this. `clock::clock_labels_languages()` returns the vector of language codes you can use.

Clock doesn't support `%OS5` or the special `%O` or `%E` prefixes, but it supports those other specifiers that stringi warns on.

``` r
x <- as.POSIXct("2022-03-18 09:20:00.1234", tz = "America/New_York")
x
#> [1] "2022-03-18 09:20:00 EDT"

# ISO Week of the year
format(x, format = "%V")
#> [1] "11"
clock::date_format(x, format = "%V")
#> [1] "11"
stringi::stri_datetime_format(x, stringi::stri_datetime_fstr("%V"))
#> Warning in stringi::stri_datetime_fstr("%V"): conversion specifier '%V' might be
#> non-portable/problematic
#> [1] "12"

# Neither clock nor stringi support this
format(x, format = "%OS6")
#> [1] "00.123399"
```

``` r
format_dt <- function(x, format, tz = "UTC", locale = NULL) {
  if (is.null(locale)) {
    format(x, format = format, tz = tz)
  } else {
    rlang::check_installed("clock")
    
    if (rlang::is_string(locale)) {
      locale <- clock::clock_locale(labels = locale)
    }
    if (!inherits(locale, "clock_locale")) {
      rlang::abort("`locale` must be a single string or a locale created by `clock::clock_locale()`.")
    }
    
    x <- clock::date_set_zone(x, zone = tz)
    
    clock::date_format(
      x = x, 
      format = format, 
      locale = locale,
      abbreviate_zone = TRUE # just to be more like format.POSIXct(x, format = "%Z")
    )
  }
}

x <- as.POSIXct("2022-03-18 09:20:00.1234", tz = "America/New_York")

format_dt(x, "%e %B")
#> [1] "18 March"

format_dt(x, "%e %B", locale = "es")
#> [1] "18 marzo"

labels <- clock::clock_labels(
  month = c("J", "F", "M", "A", "Ma", "J", "Ju", "Au", "S", "O", "N", "D"),
  weekday = clock::clock_labels_lookup("en")$weekday,
  am_pm = clock::clock_labels_lookup("en")$am_pm
)

format_dt(x, "%e %B", locale = clock::clock_locale(labels = labels))
#> [1] "18 M"
```

DavisVaughan (approved):
Implementation wise I think this looks good. My main comment is below, I'll leave that up to you

R/label-date.R
hadley: I'm torn on this, it feels more conservative to rely on stringi here, since we already use it elsewhere.

DavisVaughan: That's fine!

PR #333, opened 2022-03-17
Consistently format superseded functions

### LINKED ISSUE #308: Add lifecycle badges to retired functions
Some of the labeling functions (e.g. `percent()`) are noted as "retired". It would be nice to add lifecycle badges to them that show on top of the documentation as it's easy to miss the "Old interface" note. However, I believe this means the documentation for old interface functions would need to be separated from new ones, e.g. `label_percent()`.

hadley:
Fixes #308

@mine-cetinkaya-rundel do you mind taking a quick look through this since you asked for it? 😄 Mostly looking for copy-pasta errors.


mine-cetinkaya-rundel:
I ask for a lot of things! 😛 

These changes look good but I have a question before making a suggestion -- Should the language "This/these function/s is/are kept for backward compatibility; you should switch to ___ for new code." be included in the documentation of each superseded function? If yes, I'm happy to go through and make those suggestions inline.

hadley:
Yeah, that'd be great, thanks!

PR #335, opened 2022-03-21
Implement transformer composition

### LINKED ISSUE #287: trans, reverse, and negatives
With marine data, one typically wants to present depth on the Y-axis, starting with 0 at the top and with values of greater depth increasing as you do down the axis, as a visual analogue to descending deeper into the ocean:
image
A simple, albeit a little hacky, way of achieving this is:
`df$depth <- -df$depth`
Then scaling to one's preference, noting the negative:
`scale_y_continuous(breaks = seq(-600, 0, by = 100), limits = c(-600, 0)) +`

For various biological & oceanographic reasons, many marine species spend a disproportionate amount of time near the surface, such that there's a heterogeneous distribution of importance and data-ink in the top bit, and a lot of wasted space below.
image

Subsequently it would be nice to transform the scale and keep the reverse orientation, but, as I understand it, this isn't natively possible with the current setup:

1. Keep original depth data as positive continuous. Can use rescaling transforms (log1p, sqrt) `(trans = "log1p")` OR reverse `(trans = "reverse")`, but not both.

2. Pre-convert depth data to negative. No need to use reverse transform, but can't use (some/all) rescaling transforms because they don't work on negative values (even though [this](https://github.com/r-lib/scales/issues/214) says sqrt does due to a bug, it doesn't for me)

While I presume I could sort this with a custom `trans_new`, I suspect based on various StackOverflow posts that this is a common enough issue, in various guises, to warrant inclusion as a feature request.

Proposed solutions:

A. Add `reverse = FALSE` as a default parameter to all transforms, such that one can set this to true to rescale their cake and reverse it too (and thus remove `reverse_trans`),

B. Allow transform functions to operate on negatives by converting them to positives then converting the results back. This could be an optional parameter e.g. `allowNegative`.

Thanks in advance.

hadley:
Fixes #287 

@teunbrand how does this look to you? I implemented this without thinking too much, so I might have got the `domain` stuff wrong.


R/trans-compose.R
```diff
@@ -0,0 +1,51 @@
...
+  )
+  if (any(is.na(domain))) {
+    abort("Sequence of transformations yields invalid domain")
+  }
+  domain <- range(domain)
+
+  names <- vapply(trans_list, "[[", "name", FUN.VALUE = character(1))
+
+  trans_new(
+    paste0("composition(", paste0(names, collapse = ","), ")"),
```
teunbrand, line 31: Apologies for nitpicking, but this might print a little bit nicer with a space behind the comma, e.g.:
`paste0("composition(", paste0(names, collapse = ", "), ")")`

R/trans.r
```diff
@@ -79,13 +82,19 @@ lines.trans <- function(x, ..., xlim) {
...
+as.trans <- function(x, arg = deparse(substitute(x))) {
   if (is.trans(x)) {
-    return(x)
+    x
+  } else if (is.character(x) && length(x) >= 1) {
+    if (length(x) == 1) {
+      f <- paste0(x, "_trans")
+      match.fun(f)()
+    } else {
+      compose_trans(!!!lapply(x, as.trans))
```
teunbrand, line 93: The `as.trans` step seems to be repeated here as well as at the top of `compose_trans()`. We could probably remove this one and use the one in `compose_trans()`, unless I might be missing something.

teunbrand:
This looks great, I'm pretty excited about this!

R/trans-compose.R
hadley: I get where you're coming from, but I wanted this to look like a single "name", i.e. no spaces.

PR #339, opened 2022-03-25
New approach to variable rescaling

### LINKED ISSUE #264: label_number_si accuracy when labelling 'K' and 'M'
Hello!

I've been trying to scale my axis, that goes from 0 to 1.2M. Ideally, I would like to get numbers below 1M with K (and no decimals) but numbers above with one decimal (that is, 1.0M, 1.2M)

When I try default `accuracy`, I may get confused by numbers above 1M:
`demo_continuous(c(1, 1.2e6), label = label_number_si())`
image

But if I change accuracy, then I get decimals too for numbers below 1M:
`demo_continuous(c(1, 1.2e6), label = label_number_si(0.1))`

image

It would be awesome to have some parameters to control this kind of options.

Thanks!

hadley:
@davidchall I started getting worried that the change to `label_number_si()` was likely to affect too much existing code, and after much refactoring this is what I ended up with:

* `number()` gets a new argument called `scale_cut`, equivalent to `dollar()`'s `rescale_large`.
* `rescale_long_scale()` and `rescale_short_scale()` are renamed to `cut_short_scale()` and `cut_long_scale()`
* Add new `cut_si()` and `cut_bytes()` to pull out the code previously nested inside `label_number_si()` and `label_bytes()`. This involved refactoring `scale_cut()` to take a vector of break points, not `log10` break points.

Along the way, I also fixed #264 by refactoring `number()` to allow `accuracy` to be a vector (which then requires some manual looping over `nsmall`).

I think overall this is a cleaner separation of concerns and should be a little more flexible because it's implemented in `number()` rather than the wrappers.


davidchall:
Hi @hadley - I remember the design being quite tricky when I wrote this code a couple of years ago, so I really appreciate you taking another look at it before release. I wondered about surfacing the long and short scales in `number()`, so it's nice to see this here. I'll try to review over the weekend. Thanks!

tests/testthat/test-label-dollar.R
```diff
@@ -25,15 +25,15 @@ test_that("rescale_large works", {
   x <- 10^(seq(0, 18, 3))
```
davidchall, line 25: We need to rename the 2 testthat blocks because they reference the `rescale_large` argument.

tests/testthat/test-label-dollar.R
```diff
@@ -25,15 +25,15 @@ test_that("rescale_large works", {
...
-    label_dollar(rescale_large = rescale_long_scale())(x),
+    label_dollar(scale_cut = cut_long_scale())(x),
     c("$1", "$1K", "$1M", "$1,000M", "$1B", "$1,000B", "$1T")
   )
   expect_equal(
-    label_dollar(rescale_large = c(k = 3L, m = 6L, bn = 9L, tn = 12L))(x),
+    label_dollar(scale_cut = c(k = 1e3, m = 1e6, bn = 1e9, tn = 1e12))(x),
     c("$1", "$1k", "$1m", "$1bn", "$1tn", "$1,000tn", "$1,000,000tn")
   )
 })
```
davidchall, line 39: It might also be good to have a test when the currency symbol is a suffix (e.g. France). It seems like an edge case we should be checking for regressions.

tests/testthat/test-label-number.r
```diff
@@ -118,3 +125,43 @@ test_that("precision ignores Inf and NA", {
...
+  expect_equal(number(1e15, scale_cut = cut_short_scale()), "1 000T")
+})
+
+test_that("scale_cut checks its inputs", {
+  expect_snapshot(error = TRUE, {
+    number(1, scale_cut = 1)
+    number(1, scale_cut = "x")
+    number(1, scale_cut = c(x = 0, y = 1))
+  })
+
```
davidchall, line 166: Can we please test that `unit` is a required argument of `cut_si()`?

tests/testthat/test-label-number.r
```diff
@@ -118,3 +125,43 @@ test_that("precision ignores Inf and NA", {
...
+test_that("scale applied before scale_cut", {
+  expect_equal(number(500, scale = 2, scale_cut = cut_short_scale()), "1K")
+})
+
+test_that("cut_si() adds space before unit", {
+  expect_equal(number(c(1, 1000), scale_cut = cut_si("m")), c("1 m", "1 km"))
+})
+
+test_that("handles out-of-range inputs", {
+  expect_equal(number(1e15, scale_cut = cut_short_scale()), "1 000T")
```
davidchall, line 157: I think we should also test the low end, because `cut()` is directional.
```suggestion
  expect_equal(number(1e15, scale_cut = cut_short_scale()), "1 000T")
  expect_equal(number(1e-25, scale_cut = cut_si("")), "0.1y")
```

tests/testthat/test-label-number.r
```diff
@@ -46,6 +46,13 @@ test_that("prefix applied before negative format", {
   expect_equal(number(-1, prefix = "$", style_negative = "parens"), "($1)")
 })
 
+test_that("accuracy is vectorised", {
```
davidchall, line 49: This is a really nice bonus! 😄

tests/testthat/test-label-number.r
```diff
@@ -118,3 +125,43 @@ test_that("precision ignores Inf and NA", {
...
+})
+
+test_that("scale applied before scale_cut", {
+  expect_equal(number(500, scale = 2, scale_cut = cut_short_scale()), "1K")
+})
+
+test_that("cut_si() adds space before unit", {
+  expect_equal(number(c(1, 1000), scale_cut = cut_si("m")), c("1 m", "1 km"))
+})
+
```
davidchall, line 155: I think we should test how `cut_si("m")` handles zero. It seems like an edge case due to how `cut_si()` bakes the units into the scale suffix. IMHO the expected result should be "0 m".

tests/testthat/test-label-bytes.R
```diff
@@ -4,7 +4,7 @@ test_that("auto units always rounds down", {
 
 test_that("auto units handles 0 and other special values", {
   expect_equal(label_bytes()(NA_real_), NA_character_)
-  expect_equal(label_bytes()(0), "0 B")
+  expect_equal(label_bytes()(0), "0")
```
davidchall, line 7: IMHO we should have units after zero.

NEWS.md
```diff
@@ -35,27 +35,24 @@
...
+  `label_number(scale_cut = cut_scale_short())` or
+  `label_number(scale_cut = cut_SI("m"))`.
+
+* `number()` gains a new `scale_cut` argument that enables independent 
+  scaling of different parts of the range. This is useful in `label_dollar()` 
+  to support scaling of large numbers by suffix (e.g. "M" for million, "B" 
+  for billion). It can be used with `cut_short_scale()` when billion =
+  thousand million and `cut_long_scale()` when billion = million million.
+  Additionally, the accuracy is now computed per scale category, so rescaled
+  values can have different numbers of decimal places.
```
davidchall, line 51: I know I didn't implement the final version, but I did spend quite a bit of time thinking about this problem. Could I possibly request something like `(with help from @davidchall)`?

davidchall (commented):
I think this is absolutely fantastic! 🎉  It feels much more natural, and I'm sure it will be easier to maintain as a result. Most of my feedback was very minor.

My only remaining concern is about assigning units to the number zero. My previous solution would output "0 B" from the bytes labeler, "0 m" from the SI unit labeler, and "0" from the currency labeler (which has no unit). I think these now zeros always omit the unit. I think this has a couple of disadvantages, and I wonder if others would report this later?

1. On plot axes, it'll look weird to have inconsistent tick labels (e.g. `0, 0.5 m, 1.0 m, 1.5 m, 2.0 m`).

2. From a scientific standpoint, I think there is a difference between "0" and "0 m". The latter implies that the measured quantity has a dimension of length. To omit the unit is odd because it suggests the dimension of the axis is inconsistent (length _and_ dimensionless).

In conclusion, I can't wait to start using these functions! 😄

NEWS.md
hadley: Yes, absolutely! Many apologies for accidentally editing out your contributions to this feature.

R/label-number.r
```diff
@@ -258,3 +309,76 @@ precision <- function(x) {
     pmin(precision, 1)
   }
 }
+
+# each value of x is assigned a suffix and associated scaling factor
+scale_cut <- function(x, breaks, scale = 1, accuracy = NULL, suffix = "") {
+
+  if (!is.numeric(breaks) || !is_named(breaks)) {
+    abort("`scale_cut` must be a named integer vector")
```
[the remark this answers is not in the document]
hadley: Oh yeah, I relaxed this not because I think people will do (e.g.) `1.5` but so they can write `1` instead of `1L`.

tests/testthat/test-label-number.r
hadley: This doesn't seem worth testing to me, as it's obvious via inspection of the code, and any test would effectively just be testing that base R generates a useful error message here.

tests/testthat/test-label-number.r
hadley: Ok, should be fixed now.

NEWS.md
davidchall: No worries!

NEWS.md
```diff
@@ -35,27 +35,25 @@
...
+  for billion). It can be used with `cut_short_scale()` when billion =
+  thousand million and `cut_long_scale()` when billion = million million
+  (initial implementation provided by @davidchall). Additionally, the accuracy 
+  is now computed per scale category, so rescaled values can have different 
+  numbers of decimal places (#339).
+
 * `label_bytes()` now correctly accounts for the `scale` argument when choosing
-  auto units (@davidchall, #235).
+  auto units (@davidchall, #235) and leaves `0` as is (instead of formatting to 
+  "0 B") for consistency with `label_number_si()`.
```
davidchall, line 56: @hadley I think this addition should be removed, now that we format as "0 B".

hadley: Thanks, I'll fix

PR #341, opened 2022-04-03
Derivatives of transformations

### LINKED ISSUE #322: Feature request (and offer for help): optionally supply derivatives of transforms and inverses
Something that comes up regularly when visualizing analytical distributions on transformed scales is the need to correct the density function for the scale transformation (the so-called "Jacobian correction"), which requires the derivative of the inverse of the scale transformation.

For example, {[ggdist](https://mjskay.github.io/ggdist/)} allows visualizing analytical distributions on transformed scales and automatically determines the Jacobian in order to do this correctly. E.g. a log-normal distribution plotted on a log scale should look like a normal distribution (without this correction it would not):

``` r
library(ggplot2)
library(magrittr)
library(ggdist)

data.frame(dist = "lnorm") %>%
  ggplot(aes(y = 0, dist = dist, arg1 = log(10), arg2 = 2*log(10))) +
  stat_dist_halfeye() +
  scale_x_log10(breaks = 10^seq(-5,7, by = 2))
```

<sup>Created on 2021-12-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

In {ggdist} we currently do this first by attempting to get the symbolic derivative by applying `stats::D()` to the scale transformation function definition (this usually fails) and then fall back to numerical derivatives. Naturally, the latter are not perfect and will fail in some cases, degrading the user experience.

### The ask / offer

It occurred to me that this can't be the only application where it would be helpful to know the derivative of the `$transform()` and/or `$inverse()` of a scale transformation. So I would like to propose adding two additional fields to the `trans` objects in {scales}:

- `transform_grad` (or `grad_transform` or `d_transform` or `transform_deriv` or ...) giving the derivative of `transform`
- `inverse_grad` (or `grad_inverse` or `d_inverse` or `inverse_deriv` or ...) giving the derivative of `inverse`

These would be optional fields (may be `NULL`) that not all transformations are required to have. This would retain backwards compatibility, and would also work for my purposes (as I can continue falling back to symbolic and numerical differentiation as needed), but would (1) immediately improve the user experience with existing scale functions by supplying known derivatives in this package and (2) allow users to easily fix cases with manual scales when numerical methods fail by supplying the derivative manually.

If there is interest in something like this, I am happy to supply a PR that creates these fields and supplies derivatives for the existing transformations in `R/trans-numeric.r`. Thanks!

mjskay:
This PR closes #322 by supplying derivatives of transformation functions and inverses where possible.

Specifically, it:

- Adds two optional, may-be-`NULL` fields to transform objects: `$d_transform` and `$d_inverse`. These functions are the derivatives of `$transform` and `$inverse`.
- Supplies derivatives for all existing numeric transformations
- Applies the chain rule to give derivatives for `compose_trans()` (unless any derivatives in the transformation are missing, in which case the corresponding derivative function is `NULL`).
- Includes tests for all of the above.

Some possible discussion points:

- I initially used `$transform_deriv` and `$inverse_deriv`, but found that existing tests used partial matching on `$trans` and `$inv`, so these names caused failures. I wasn't sure if other legacy code relies on this partial matching, so I conservatively went with the names `$d_transform` and `$d_inverse`. I am not sure these are the best names and am happy to change them.
- I did not supply derivatives for the date transform. Arguably those derivatives could be a constant like `identity_trans()`, at least for applications like I am imagining these will be used for (plotting densities), but I went with the more conservative choice of not including it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,6 +23,8 @@
 * Add an inverse (area) hyperbolic sine transformation `asinh_trans()`, which
   provides a logarithm-like transformation of a space, but which accommodates
   negative values (#297)
+* Transformation objects can optionally include the derivatives of the transform
+  and the inverse transform (@mjskay, #322).
 
 # scales 1.2.1
 
--- a/R/trans-compose.R
+++ b/R/trans-compose.R
@@ -27,11 +27,16 @@ compose_trans <- function(...) {
 
   names <- vapply(trans_list, "[[", "name", FUN.VALUE = character(1))
 
+  has_d_transform <- all(lengths(lapply(trans_list, "[[", "d_transform")) > 0)
+  has_d_inverse <- all(lengths(lapply(trans_list, "[[", "d_inverse")) > 0)
+
   trans_new(
     paste0("composition(", paste0(names, collapse = ","), ")"),
-    transform = function(x) compose_fwd(x, trans_list),
-    inverse   = function(x) compose_rev(x, trans_list),
-    breaks    = function(x) trans_list[[1]]$breaks(x),
+    transform   = function(x) compose_fwd(x, trans_list),
+    inverse     = function(x) compose_rev(x, trans_list),
+    d_transform = if (has_d_transform) function(x) compose_deriv_fwd(x, trans_list),
+    d_inverse   = if (has_d_inverse) function(x) compose_deriv_rev(x, trans_list),
+    breaks      = function(x) trans_list[[1]]$breaks(x),
     domain = domain
   )
 }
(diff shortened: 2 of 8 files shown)


thomasp85:
@mjskay would you update the PR to remove conflicts?

teunbrand:
I'd like to point out a small interaction with #360 here.
That PR adds the inverse hyperbolic sine transformation.
I believe those derivatives should be `1 / sqrt(x^2 + 1)` for the transformation and `cosh()` for the inverse.
I'm not sure what gets merged first or which PR should implement this (if any), but it might be neat to include.

pearsonca:
if this goes in first, happy to update #360 to use this feature (and double check derivates etc)

mjskay:
Conflicts resolved here.

Thanks @teunbrand @pearsonca. Similarly, if #360 goes first I'd be happy to add those derivatives here; either way works for me.

thomasp85:
We'll do #360 first. Will let you know once merged

mjskay:
@pearsonca I think @thomasp85 was suggesting to merge #360 without the derivatives and then deal with derivatives here, which probably makes sense in case any feedback on this PR causes changes to the derivative stuff (e.g., name changes or what have you).

thomasp85:
#360 has been merged - can you update this PR with derivatives for that and then we can get this in

mjskay:
Done!

Also thanks for all the hard work shepherding through all these issues and PRs :).

thomasp85:
Thank you for all the work

Issue #347, opened 2022-04-19
Snapshot test is locale dependent


johanneskoch94:
Hi, 
testing fails on my machine, due to a different locale that prints out "Okt" instead of "Oct", the latter of which is expected in the snapshot. This concerns line 34 and 41 specifically. I guess `withr::local_locale()` would do the trick, but since it's not being imported I'm not sure what the best solution is. Just wanted to let you know. Thanks for all the amazing packages!
https://github.com/r-lib/scales/blob/a2552d2aa1f04e1b394a7f7a23ecda92570a5971/tests/testthat/test-label-date.R#L27-L44

mjskay:
FWIW {testthat} imports {withr}, so adding {withr} to Suggests in {scales} wouldn't add a dependency that's not already downstream.

PR #348, opened 2022-04-19
Fix typos

johanneskoch94:
This PR fixes 2 typos in the README. Testing fails (locally at least) due to #347.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -40,7 +40,7 @@ devtools::install_github("r-lib/scales")
 
 ### Breaks and labels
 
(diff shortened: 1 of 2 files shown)


johanneskoch94:
So the checks are successful (#347 isn't relevant when testing via Github actions), but pkgdown fails and I don't understand why, I'm sorry. (pkgdown::build_site() is successful locally.).

johanneskoch94:
So the checks are successful (#347 isn't relevant when testing via Github actions), but pkgdown fails and I don't understand why, I'm sorry. (pkgdown::build_site() is successful locally.).

thomasp85:
Thanks! ❤️

PR #354, opened 2022-06-10
Fix #352 to make label_date_short() render all labels of the same level inline

### LINKED ISSUE #352: In label_date_short, can all days, months and years be lined up (regardless of the vjust selected)?
`scale_date_short` would look better if all the date labels of the same level lined up.

So in the example below, the "Oct" and "Apr" labels would be in line with the "Jan" and "Jul" labels
image

davidhodge931:
Before with `theme(axis_text = element_text(vjust = 0.5)`:
image

After with `theme(axis_text = element_text(vjust = 0.5)`:
image

Haven't done a pull request to another library before, so apologies if I've done something incorrect

### DIFF
--- a/R/label-date.R
+++ b/R/label-date.R
@@ -81,9 +81,9 @@ label_date_short <- function(format = c("%Y", "%b", "%d", "%H:%M"), sep = "\n") 
     }
 
     for_mat <- cbind(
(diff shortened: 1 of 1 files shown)


davidhodge931:
Not required, as issue seems to be fixed somehow

PR #359, opened 2022-08-25
set default domain for `probability_trans`

pearsonca:
The domain of a q... distribution function is a probability, aka (0,1)

Alternatively, sensible to exclude domain from the arguments, and simply set it in `trans_new` without allowing user intervention.


thomasp85:
Fixed in another PR - thanks for the contribution nevertheless!

Issue #297, opened 2020-10-07
Add inverse hyperbolic sine transformation


frederikziebell:
Sometimes it's nice to have a 'log'-transformation that also works for negative values to have points more evenly spread-out:

```
ihs <- function(x){log(x+sqrt(x^2+1))}
ihs_trans <- scales::trans_new(name="ihs",transform=ihs, inverse=sinh)

set.seed(1)
data <- tibble(
  x=runif(10^4),
  y=rnorm(10^4)
)

data %>% 
  ggplot(aes(x,y)) +
    geom_point(alpha=.7) +
    ggtitle("regular")

data %>% 
  ggplot(aes(x,y)) +
    geom_point(alpha=.7) +
    scale_y_continuous(trans=ihs_trans) +
    ggtitle("ihs")
```

![image](https://user-images.githubusercontent.com/28304028/95364269-eac60d80-08d0-11eb-887b-490f5f55c1bd.png)
![image](https://user-images.githubusercontent.com/28304028/95364289-f1548500-08d0-11eb-8fef-268d152e93fd.png)

thomasp85:
Fixed by #360

PR #360, opened 2022-08-25
add inverse hyperbolic sin transform

pearsonca:
Saw #297, and seemed easy enough to fix while I was looking at that area anyway.


thomasp85:
Would you like to update the PR to remove merge conflicts

pearsonca:
Happy to - anything else I should address while in there?

thomasp85:
No, thanks. Better to keep the PRs lean and focused

thomasp85:
However, can I get you to add a bullet in NEWS about the new feature

pearsonca:
Can do re NEWS.

Question - I've written merge conflict resolution on `probability_trans` as moving the `domain` default setting to the signature. The existing resolution is setting it internally - what's the style preference here? I'd lean towards making it visible to the user via signature (and also allowing for some clever hacks with user defined q/p functions), but am also sympathetic to the-mathematical-definition-of-a-probability-is-...

tests/testthat/test-trans-numeric.R
```diff
@@ -116,3 +116,8 @@ test_that("Yeo-Johnson transform works", {
   expect_equal(yj_trans(lambdas[2])$transform(x[[2]]), expected_data[[2]])
   expect_equal(yj_trans(lambdas[3])$transform(x[[3]]), expected_data[[3]])
 })
+
+test_that("probability transforms have domain (0,1)", {
+  expect_true(all(logit_trans()$domain == c(0, 1)))
+  expect_true(all(probit_trans()$domain == c(0, 1)))
```
olivroy, line 122: I think this is not an accurate test.
A better one could be 
```r
  expect_true(all(logit_trans()$domain %in% c(0, 1)))
  expect_true(all(probit_trans()$domain %in% c(0, 1)))
```

Or if you always expect a length 2, exactly `c(0, 1)`
```r
  expect_equal(logit_trans()$domain, c(0, 1))
```

Or if you are not sure of the order

```r
  expect_setequal(logit_trans()$domain, c(0, 1))
```

thomasp85:
I think the domain should be fixed and set internally rather than as part of the construction

tests/testthat/test-trans-numeric.R
mjskay: Isn't this testing that the domain is exactly `c(0, 1)`? Then it should be `expect_equal(logit_trans()$domain, c(0, 1))` or `expect_identical(logit_trans()$domain, c(0, 1))`

olivroy: I think you are right. 
I wasn't exactly sure, but I knew that  `expect_true(all(logit_trans()$domain == c(0, 1)))`

pearsonca: e.g.
```r
c(1, 1) == c(2, 1)
```
...yields
```r
FALSE TRUE
```
...a logical vector, which when tested with `all()` would yield `FALSE` (in this case). It is equivalent to `expect_equal`.

So not sure what "not an accurate test" means? I agree that the `expect_equal` syntax is preferrable, so will make that change.

I think the length test would be more appropriate as a general test of all transform generators.

olivroy: It is just that `all()` doesn't handle edge cases very well.
```r
all(integer(0) == c(0, 1)) |> testthat::expect_true()
```

This test passes.

pearsonca: to be clear, the explicit intent here is that the domain values lower == 0 & upper == 1, which is equivalent to all(c(lower, upper) == c(0, 1)). this isn't an %in% relationship

olivroy: Ok, the correct test would therefore be 

```r
expect_equal(logit_trans()$domain, c(0, 1))
```

Because in the (unlikely) scenario that
`logit_trans()$domain = integer(0)`, the current test would pass even though it is incorrect.

pearsonca: The key defect in the all approach seems to be the "empty set" case, so that aspect of it actually does invalidate the test.

pearsonca: also: just now seeing replies. not a crazy person ignoring your points.

olivroy: No problem. Sorry. I wanted to make sure I was clear because I saw this exact problem in another repo that hid a bug silently for a while. :) 

```r
  expect_equal(logit_trans()$domain, c(0, 1))
```
note the `,` instead of `==`

tests/testthat/test-trans-numeric.R
```diff
@@ -116,3 +116,8 @@ test_that("Yeo-Johnson transform works", {
   expect_equal(yj_trans(lambdas[2])$transform(x[[2]]), expected_data[[2]])
   expect_equal(yj_trans(lambdas[3])$transform(x[[3]]), expected_data[[3]])
 })
+
+test_that("probability transforms have domain (0,1)", {
+  expect_equal(logit_trans()$domain == c(0, 1))
+  expect_equal(probit_trans()$domain == c(0, 1))
+})
```
thomasp85, line 123: `expect_equal()` requires two arguments, not a single comparison

thomasp85:
Also, can you remove the derivative part - I rather we add that all together in one PR

tests/testthat/test-trans-numeric.R
pearsonca: both done!

thomasp85:
Thank you 🙏

Issue #356, opened 2022-07-06
Add `style_positive` and `style_negative` to other continuous labelers


bwiernik:
I most often work with numbers that are changes in percentages (e.g., -2%, +5%). I mostly use `label_percent()` for this. However, I'd like to be able to use the `style_positive` and `style_negative` arguments from `label_number()` to format with proper plus and minus signs. Would it be possible to add these arguments to other continuous labelers (especially `label_percent()`)?

factorialmap:
library(scales)

demo_continuous(c(-50,50),
                labels = label_percent(scale = 1,
                                       style_positive = "plus",
                                       style_negative = "minus")
                )


hope this help

bwiernik:
Ah, so the issue is just that they aren't documented

pearsonca:
@bwiernik i've proposed a revision that addresses this, by changing where `...` points to (namely, `label_numeric`, which includes the `style_(positive|negative)` documentation - is sufficient change?

bwiernik:
That should work

PR #361, opened 2022-08-25
update label_percent ... documentation

pearsonca:
addresses #356: reflects pass through `label_number` first (i.e. can also include style_+/-)

also update internals of label_percent to skip superceded function.

### DIFF
--- a/R/label-percent.R
+++ b/R/label-percent.R
@@ -1,5 +1,6 @@
 #' Label percentages (2.5%, 50%, etc)
 #'
+#' @param ... Other arguments passed on to [label_number()]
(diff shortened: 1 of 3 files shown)


teunbrand:
I'm not a maintainer of this package, but it seems like this is one of those use cases where `@inheritDotParams` would shine. For more information, see [this paragraph](https://r-pkgs.org/man.html#inheriting-documentation) of the R packages book.

pearsonca:
Hmm, agree. Wait one.

thomasp85:
Fixed outside of this issue. However, thanks for looking into it! ❤️

PR #374, opened 2022-12-06
`train_continuous()` coerces `new` to `numeric`

### LINKED ISSUE #369: Bug: `train_continuous()` with integer64
Currently, `train_continuous()` gives nonsensical ranges for integer64 input.

``` r
library(scales)

new <- bit64::as.integer64(1:10)

train_continuous(new)
#> [1] 4.940656e-324 4.940656e-323
```

Looking at the the `base::range.default` function, it seems that concatenating NULL with an integer64 vector causes weird casting behaviour. This doesn't happen using the `vctrs::vec_c()` function for example.

``` r
existing <- NULL

c(existing, new)
#>  [1] 4.940656e-324 9.881313e-324 1.482197e-323 1.976263e-323 2.470328e-323
#>  [6] 2.964394e-323 3.458460e-323 3.952525e-323 4.446591e-323 4.940656e-323

vctrs::vec_c(existing, new)
#> integer64
#>  [1] 1  2  3  4  5  6  7  8  9  10
```

<sup>Created on 2022-11-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

There are a few options to prevent this. In the following line https://github.com/r-lib/scales/blob/f7056ad39f4e476e1288082630d59084a41ee635/R/scale-continuous.r#L49

The input can be pre-combined:

``` r
suppressWarnings(range(vctrs::vec_c(existing, new), na.rm = TRUE, finite = TRUE))
```

The input can be coerced to numeric:

``` r
suppressWarnings(range(existing, as.numeric(new), na.rm = TRUE, finite = TRUE))
```

Or can be branched depending on the 'null'-ness of `existing`.

``` r
if (is.null(existing)) {
  suppressWarnings(range(new, na.rm = TRUE, finite = TRUE))
} else {
  suppressWarnings(range(existing, new, na.rm = TRUE, finite = TRUE))
}
```

teunbrand:
This PR aims to fix #369 and fix https://github.com/tidyverse/ggplot2/issues/2377.

I chose to coerce instead of the two alternatives in #369, because of this expectation:

https://github.com/r-lib/scales/blob/f7056ad39f4e476e1288082630d59084a41ee635/R/scale-continuous.r#L36

The reprex from the ggplot2 issue should now work:

``` r
#> ℹ Loading scales
library(ggplot2)

df <- data.frame(x = bit64::as.integer64(1:10))
ggplot(df) + geom_histogram(aes(x))
#> Don't know how to automatically pick scale for object of type <integer64>.
#> Defaulting to continuous.
#> `stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
```

<sup>Created on 2022-12-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


teunbrand:
Done.

PR #375, opened 2022-12-12
Update 'domain' documentation

teunbrand:
This PR proposes an extra line of documentation for the 'domain' argument in `trans_new()`.

Briefly, since ggplot2 3.4.0, breaks are computed on the expanded limits capped by the domain, which requires it to be transformable. This requirement has not been communicated very clearly, as encountered in https://github.com/tidyverse/ggplot2/issues/5105, that this PR aims to amend.

If so desired (but not yet included in this PR), the transformability of the domain could also be tested in `trans_new()`, which could then emit a warning if the domain is untransformable.


thomasp85:
Can you update this as well?

teunbrand:
Done.

PR #394, opened 2023-07-25
Report domains for several transformations

teunbrand:
Briefly, ggplot2 uses the domain to figure out in what range it can ask a transformation to compute breaks after scale expansion.
Having a default domain of `c(-Inf, Inf)` works for most transformations, but not all.
In particular the `boxcox_trans()` doesn't really work as intended as a consequence of having the default domain. While trying to fix that, I decided to also add domains for `asn_trans()` and probability transformations.

A short demo of the issue and proposed fix:

``` r
library(scales)

# Note: I'm not giving the scale negative values, some become negative through expansion
demo_continuous(1:10, trans = boxcox_trans(2))
#> scale_x_continuous(trans = boxcox_trans(2))
#> Error: boxcox_trans must be given only positive values. Consider using modulus_trans instead?

# With this PR:
#> ℹ Loading scales

demo_continuous(1:10, trans = boxcox_trans(2))
#> scale_x_continuous(trans = boxcox_trans(2))
```

<sup>Created on 2023-07-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


thomasp85:
Can you update to remove conflict?

teunbrand:
Certainly. Also realised that `log1p_trans()` should have a domain.

teunbrand:
Failure on devel seems related to some unicode <-> grid interaction? Seems unrelated to the 3 PRs you asked me to update.

thomasp85:
Yeah, don't worry about that

PR #397, opened 2023-09-06
Transformation suffixes to prefixes

### LINKED ISSUE #390: `*_trans` suffixed functions would be easier to use if they were renamed to be `trans_*` prefixed functions
A trans prefix would work well with the Rstudio auto-complete making it easier to search through the functions.

It would also be consistent with `scales::oob_*` functions, `ggplot2::stat_*` functions, and `ggplot2::position_*` functions etc.

teunbrand:
This PR aims to fix #390.

Briefly, it renames all `*_trans` functions to `trans_*` functions, while keeping the suffixed versions as aliases.
The `as.trans()` function first tries the prefixed version, and when unavailable, tries the suffixed version for backward compatibility.


thomasp85:
If you are up for it I'd like to rename `trans` to `transform` in the API. This could be well folded into this PR

teunbrand:
Sure. Just to clarify, we then should end up with the `transform_log10()` function and for backward compatibility also the `log10_trans()` function, right?

teunbrand:
I'm going to make a new PR for that, as git merge just wipes all of this PRs changes and I'm in no mood to manually go through every change. I'll start with a clean slate

Issue #393, opened 2023-07-22
`compose_trans` does not work where "reverse" is last


davidhodge931:
``` r
library(scales)
#> Warning: package 'scales' was built under R version 4.3.1
demo_continuous(10^c(-2:4), trans = c("reverse", "log10"))
#> scale_x_continuous(trans = c("reverse", "log10"))
#> Error in `compose_trans()`:
#> ! Sequence of transformations yields invalid domain
#> Backtrace:
#>     ▆
#>  1. └─scales::demo_continuous(10^c(-2:4), trans = c("reverse", "log10"))
#>  2.   └─scales:::demo_ggplot(x, "scale_x_continuous", ...)
#>  3.     └─ggplot2 (local) scale(NULL, ...)
#>  4.       └─ggplot2::continuous_scale(...)
#>  5.         └─scales::as.trans(trans)
#>  6.           └─scales::compose_trans(!!!x)
#>  7.             └─rlang::abort("Sequence of transformations yields invalid domain")
```

<sup>Created on 2023-07-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

teunbrand:
This behaviour makes sense because the reverse operation transforms the data by multiplying by `-1` and you cannot take the log of negative numbers. The error that the domain has become invalid is appropriate I think.

davidhodge931:
Maybe a sentence in help to tell people?

PR #408, opened 2023-11-06
Correct domain calculation for compose_trans

mjskay:
This PR corrects the domain calculations for `compose_trans()`, which appear to be incorrect. 

## The problem

Currently, `compose_trans()` pushes the domain of the first function through all transformations to calculate the domain. This procedure, when it works (i.e. if the range of each transformation in the chain is contained within the domain of the next one), yields the *range* of the composed set of transformations, not the domain.

This leads to incorrect domains; e.g.:

```r
compose_trans("sqrt", "reverse")$domain
## [1] -Inf    0
```

The correct domain here is just the domain of `sqrt()`, which is `c(0, Inf)`. We can see this by plotting it:

```r
plot(compose_trans("sqrt", "reverse"), xlim = c(-1,4))
## Warning message:
## In trans$transform(x) : NaNs produced
```
image

This also leads to `compose_trans()` incorrectly erroring on valid compositions; e.g.:

```r
compose_trans("reverse", "log")
## Error in `compose_trans()`:
## ! Sequence of transformations yields invalid domain
## Run `rlang::last_trace()` to see where the error occurred.
```

Yet, this should be a valid transformation, just with the domain `c(-Inf, 0)` instead of the original domain of `c(0, Inf)` (this also offers a solution to the question about reverse + log in #393: the transformation is valid so long as the domain is correctly calculated).

## The solution

This PR adjusts the domain calculation to push the first domain forward through the transformations, intersecting it with the domains of each transformation along the way, then pushing the resulting range back through the inverses to get the domain.

On the above two examples, with this PR, we have:

```r
compose_trans("sqrt", "reverse")$domain
## [1]   0 Inf
```

and

```r
compose_trans("reverse", "log")$domain
## [1]    -Inf -1e-100
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,6 +23,7 @@
 * Add an inverse (area) hyperbolic sine transformation `asinh_trans()`, which
   provides a logarithm-like transformation of a space, but which accommodates
   negative values (#297)
+* Correct the domain calculation for `compose_trans()` (@mjskay, #408).
 
 # scales 1.2.1
 
--- a/R/trans-compose.R
+++ b/R/trans-compose.R
@@ -16,10 +16,22 @@ compose_trans <- function(...) {
     cli::cli_abort("{.fun compose_trans} must include at least 1 transformer to compose")
   }
 
-  # Resolve domains
-  suppressWarnings(
-    domain <- compose_fwd(trans_list[[1]]$domain, trans_list[-1])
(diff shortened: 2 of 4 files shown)


thomasp85:
Thanks - merging your previous PR introduced a conflict. Can I get you to resolve that?

PR #410, opened 2023-11-06
Transformation suffixes to prefixes (attempt 2)

### LINKED ISSUE #390: `*_trans` suffixed functions would be easier to use if they were renamed to be `trans_*` prefixed functions

teunbrand:
This PR aims to fix #390. It supplants #397.

It does the following:
* It renames all `*_trans` suffixed functions to `transform_*` prefixed functions. The suffixed versions are kept as an alias.
* Changed the class of `"trans"` objects  to `"transform"` class objects.
* `as.transform()`, previously `as.trans()` has mechanism to look for the prefixed names first, then tries to find suffixed versions.

I was unsure about if `trans_new()` or `trans_range()` should be renamed. As they aren't transformations, I didn't rename them. I'd be happy to rename them if so desired.


thomasp85:
Let's rename it all :-) but keep the old around

teunbrand:
Right, but if we call it `transform_new()` and `transform_range()`, folks might expect that these are also transformations to be passed to scales. How do you feel about `new_transform()` and `range_transform()`?

thomasp85:
`trans_new()` should become `new_transform()` so it is in line with how constructors are generally named now

`trans_range()` should probably be renamed to something like `trim_to_domain()` or something

teunbrand:
Alright, should be OK now

thomasp85:
Thanks!

k-motwani:
> Let's rename it all :-) but keep the old around

It seems like the old were not kept around which causes issues downstream. Could this be remedied?

teunbrand:
Do you have a minimal reproducible example of what previously worked fine and is giving issues now?

PR #414, opened 2023-12-14
Confusion between y and lambda in the documentation of the Box-Cox transformation 

EricMarcon:
The documentation of the Box-Cox transformation is not correct. It confuses the parameters $y$ and $\lambda$.
This PR corrects it.

To be convinced, read the equations in the Details section. $y$ may take any value but the fractions are 0/0 when $\lambda = 0$.


R/transform-numeric.R
```diff
@@ -69,14 +69,14 @@ asinh_trans <- transform_asinh
 #' The Box-Cox power transformation (type 1) requires strictly positive values and
 #' takes the following form for `y > 0`:
```
thomasp85, line 70:
```suggestion
#' takes the following form for \eqn{\lambda > 0}:
```

thomasp85:
\document

thomasp85:
Thanks

PR #416, opened 2023-12-22
Propagate call for error messages

### LINKED ISSUE #379: Improve messaging in Range R6 classes
This suggestion is lifted over from https://github.com/tidyverse/ggplot2/issues/4258.

Briefly, `train_discrete()` and `train_continuous()` throw errors that can be hard to debug in some cases.
If https://github.com/tidyverse/ggplot2/pull/5086 is merged, that would leave responsibility over Range classes with {scales}, so that is why I'm opening this issue here.

My proposal would be that the Range classes get an optional field with a scale name or aesthetic name that can be used for communicating more informative errors. Responsibility for providing an informative field would then bounce the ball back to ggplot2.

teunbrand:
This PR aims to fix #379.

Briefly, it adds a `call` argument to range training that is used to communicate error messages.
Short demo:

``` r
#> ℹ Loading scales

i_train_ranges <- function(x) {
  range <- ContinuousRange$new()
  range$train(x, call = current_call())
  range$range
}

i_train_ranges(1:10)
#> [1]  1 10
i_train_ranges(c("foo", "bar"))
#> Error in `i_train_ranges()`:
#> ! Discrete value supplied to a continuous scale
```

<sup>Created on 2023-12-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/range.R
+++ b/R/range.R
@@ -18,9 +18,10 @@ DiscreteRange <- R6::R6Class(
   inherit = Range,
   list(
     factor = NULL,
-    train = function(x, drop = FALSE, na.rm = FALSE) {
+    train = function(x, drop = FALSE, na.rm = FALSE, call = caller_env()) {
       self$factor <- self$factor %||% is.factor(x)
-      self$range <- train_discrete(x, self$range, drop, na.rm, self$factor)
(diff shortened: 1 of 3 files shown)


thomasp85 (approved):
LGTM

PR #427, opened 2024-04-16
Classed palettes with properties

### LINKED ISSUE #426: Feature request: properties for palettes
This issue follows up from https://github.com/tidyverse/ggplot2/issues/5843#issuecomment-2056499317.
Currently, palettes are function factories that just return other functions that e.g. scales in ggplot2 uses. It would be useful is we could query properties of the produced functions. However, these functions are hard to interrogate as you'd have to try to figure out properties from the closure environment. It would be useful if we could query the following properties in a standardised fashion:

* Is a palette continuous? I.e. does it accept a vector of (0, 1) input?
* Is a palette discrete? I.e. does it accept a scalar `n` as input?
* If a palette is discrete, what is the maximum supported `n`?
* Do they have build-in support for `NA` translation. This is also a subject in #372.

In addition, it would be good to have a translator function that can accept discrete palettes and turn it into continuous palettes and vice versa. We can't currently do this systematically because we don't know the maximum supported `n` for every palette.

teunbrand:
This PR aims to fix #426.

Briefly, it introduces two classes of palettes: `<pal_continuous>` and `<pal_discrete>`, which I mostly see as making {scales}' palettes more formal.

The classes are simply functions with attributes that provide information about the palette.
The `<pal_discrete>` class accepts a scalar `n` as input and spits out a vector of `n` distinct values.
The `<pal_continuous>` class accepts a numeric vector between (0, 1) and spits out a vector of equal length.
All existing palettes have been given appropriate classes.

The `palette_type()` function probes the `type` attribute, indicating what palettes can be used for, like `"colour"` or `"linetype"`.
The `palette_na_safe()` function probes `<pal_continuous>` on whether `NA` are translated (e.g. in using the `colour_ramp(na.colour)` argument).
The `palette_nlevels()` function probes `<pal_discrete>` what the maximum `n` is.

The `as_continuous_pal()` function can cast `<pal_discrete>` to `<pal_continuous>`. For colour types, this is wrapping the discrete palette in `colour_ramp()`.
The `as_discrete_pal()` function can cast `<pal_continuous>` to `<pal_discrete>`. This is mostly `palette_fun(seq(0, 1, length.out = n)`.

A small demo:

``` r
#> ℹ Loading scales

# Probing properties
pal <- pal_brewer("qual")
is_discrete_pal(pal)
#> [1] TRUE
palette_nlevels(pal)
#> [1] 8

# Converting from continuous to discrete
pal <- pal_gradient_n(rainbow(10))
is_continuous_pal(pal)
#> [1] TRUE
disc <- as_discrete_pal(pal)
is_discrete_pal(disc)
#> [1] TRUE
show_col(disc(9))
```

<sup>Created on 2024-04-16 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


teunbrand:
@thomasp85 can I put this on your radar for review sometime?

thomasp85 (approved):
LGTM

PR #447, opened 2024-09-03
fix typo in documentation for `label_pvalue()`

d-morrison:
The documentation of the `prefix` argument appears to be opposite of the actual code; here's a PR to fix.


thomasp85 (approved):
Thanks

PR #448, opened 2024-09-19
Named palettes

### LINKED ISSUE #396: Feature request: palette registration
I think it would be great if there was some mechanism in which packages can register a colour palette that then could natively interface with scales/ggplot2. It would save having to write separate `scale_colour/fill_*()` functions for packages providing the palettes, and save having to write `scale_{custom_colour_aesthetic}_{palette_type}` for packages that extend layers with custom colour-like aesthetics.

From a palette package POV you might then call

```r
scales::register_palette(name = "foobar", type = "qualitative", fun = my_palette_fun)
```

This then gets stored in a list from which one could retrieve

```r
scales::pal_registered(type = "qualitative", name = "foobar")
```

Which then gets used in e.g. ggplot2 scales. A nicety that can be added is to pre-populate these list of registered palettes with R's build-in palettes.

teunbrand:
This PR aims to fix #396.

Briefly, it adds a register of known palettes.

In some more detail, you can get one of the known palettes with `get_palette()`, set a new known palette with `set_pallette()`, see all known palettes with `palette_names()` and restore the default palettes with `reset_palettes()`. Along with these functions, there is integration with `as_continuous/discrete_pal()` and `show_col()` to work with known palettes.

The nice thing about having known palettes, is that you can refer to them by name.

``` r
#> ℹ Loading scales
show_col("Okabe-Ito")
```

This PR would nicely integrate with https://github.com/tidyverse/ggplot2/pull/6112, as it would use `as_continuous/discrete_pal()` in the default colour scales.

``` r
pal <- as_discrete_pal("magma")
show_col(pal(16))
```

Palette packages are then able to register their own palettes, which you could then also refer to by name.

``` r
cols <- c("palegreen", "deepskyblue", "magenta")
set_palette("aurora", palette = cols)
show_col(as_continuous_pal("aurora"), labels = FALSE)
```

<sup>Created on 2024-09-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


teunbrand:
Currently fails because `grDevices::palette.pals()` didn't exist in R 3.6. The aim was to continue supporting R 3.6 for a while longer, no?

davidhodge931:
Awesome work - this will be super useful! I wonder if the `set_palette` function here should instead be called `register_palette` @teunbrand? This stuff is subjective. But I would expect setting a palette would change the colours by default for subsequent colour/fill scales

thomasp85 (approved):
Looks good - thanks

PR #450, opened 2024-09-26
Update warning message for shape overuse

Aehmlo:
I accidentally transposed a couple of aesthetic mappings while constructing a plot yesterday and was (rightfully) warned about my request for 25 different shapes. However, I noticed that the wording of the warning was a little strange/confusing (emphasis mine):

> you have requested 25 values. Consider specifying shapes manually if you need **that many have them**.

It looks like this happened unintentionally in 45ca806b: the message originally read "if you must have them." This PR rewords it to "if you need that many of them," but I'm not attached to this particular phrasing.


thomasp85 (approved):
Thank you

PR #452, opened 2024-10-02
Minor break function for log10 ticks

teunbrand:
This PR aims to fix https://github.com/tidyverse/ggplot2/issues/6016.

It adds a `minor_breaks_log()` function that mirrors the calculation in `ggplot2::guide_axis_logticks()`.

By default, it tries to tone down the detailing as we have more powers of 10 covered.
In the example below, the y-axis has ~4 powers and every multiple of a power is marked.
However, in the x-axis spans ~7 powers and only every 5th multiple of a power, and the power itself is marked.

``` r
#> ℹ Loading scales
library(ggplot2)

ggplot(msleep, aes(bodywt, brainwt)) +
  geom_point() +
  scale_x_log10(minor_breaks = minor_breaks_log()) +
  scale_y_log10(minor_breaks = minor_breaks_log())
#> Warning: Removed 27 rows containing missing values or values outside the scale range
#> (`geom_point()`).
```

<sup>Created on 2024-10-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


tests/testthat/test-breaks-log.R
```diff
@@ -90,3 +90,32 @@ test_that("breaks_log with very small ranges fall back to extended_breaks", {
...
+
+test_that("minor_breaks_log rejects invalid arguments", {
+  expect_error(
+    minor_breaks_log(7),
+    "must be one of 1, 5 or 10"
+  )
+  expect_error(
+    minor_breaks_log(smallest = 0),
+    "must be a finite, positive, non-zero number."
+  )
```
thomasp85, line 120: We generally want to use `expect_snapshot_error()` unless there is a very good reason not to

thomasp85 (changes requested):

PR #453, opened 2024-10-03
Number options

### LINKED ISSUE #376: Feature: global options for decimal.mark and big.mark
Hi,

would it be relevant to have a way to globally define (at the beginning of a script) global options regarding decimal and big marks, that then will be taken into account by all `label_*()` functions if nothing else is specified?

It could simplify the writing of statistical reports on other languages than English.

Best regards

teunbrand:
This PR aims to fix #376.

Briefly, it adds a new function `number_options()` that globally sets options for formatting numbers.
Subsequently, the `label_*()` functions use these options by default.

Rendered examples of the documentation:

``` r
#> ℹ Loading scales

# Default number formatting
x <- c(0.1, 1, 1000)
label_number()(x)
#> [1] "0.1"     "1.0"     "1 000.0"

# Now again with new options set
number_options(style_positive = "plus", decimal.mark = ",")
label_number()(x)
#> [1] "+0,1"     "+1,0"     "+1 000,0"

# The options are the argument names with a 'scales.'-prefix
options("scales.style_positive")
#> $scales.style_positive
#> [1] "plus"

# Resetting the options to their defaults
number_options()
label_number()(x)
#> [1] "0.1"     "1.0"     "1 000.0"
```

<sup>Created on 2024-10-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (commented):
Just to make sure. The prior behaviour when passing in `NULL` to all these arguments was to error, right? We are not breaking any expected behaviour here

teunbrand:
In current scales, using `prefix = NULL` or `suffix = NULL` doesn't throw errors, it just doesn't add a prefix/suffix.
The `big.mark`, `decimal.mark`, `style_positive` and `style_negative` options did not accept NULL.

So I guess technically this PR might break `prefix`/`suffix`, but only if people go out of their way to write more verbose code than necessary.

If this is a problem, we can always make the argument a specific `getOption()`.

thomasp85:
I think this is probably fine

PR #454, opened 2024-10-04
Improve exponential breaks

### LINKED ISSUE #405: Better breaks calculations for exponential transformations
Superseedes #59

While we no longer see errors in ggplot2 for these transformations (the report in the original issue), the placement of breaks with various transformations still leave a lot to be desired

```r
library(ggplot2)
dat <- data.frame(x = 5:100)

ggplot(dat) + 
  aes(x = x, y = x) + 
  geom_point() +
  scale_y_continuous(trans = "exp")
```

teunbrand:
This PR aims to fix #405.

Briefly, it introduces a new breaks function for the exponential transformation: `breaks_exp()`.
It is also incorporated in `transform_exp()` as the breaks function.

While it is not perfect, it most definitely is an improvement over the old situation. The way I came up with this break algorithm is just by eyeballing it. I found that if the range is greater than 3, the exponential transformation began to deform the space by a lot. When the range is smaller than 3, it wasn't too bad, so we're still using `breaks_extended()` there. When the range is larger than 3, we're labelling 0 and the last `n - 1` integers of the range.

Some examples:

``` r
#> ℹ Loading scales

demo_continuous(c(0, 100), transform = "exp")
#> scale_x_continuous(transform = "exp")
```

Small range reverts to `breaks_extended()`:

``` r
demo_continuous(c(0, 2.5), transform = "exp")
#> scale_x_continuous(transform = "exp")
```

Just after the cutoff point:

``` r
demo_continuous(c(0, 3.5), transform = "exp")
#> scale_x_continuous(transform = "exp")
```

Even though the limits don't include 0, the expanded range does, so we still display 0 here:

``` r
demo_continuous(c(100, 104), transform = "exp")
#> scale_x_continuous(transform = "exp")
```

<sup>Created on 2024-10-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85 (approved):
Thanks for this. We can always improve further if cases with bad results are reported

PR #459, opened 2024-10-07
New `label_glue()` function

### LINKED ISSUE #457: Feature request: `label_glue()`
So you can do `scale_y_continuous(label = label_glue("Hi {x}!"))` and it will produce `Hi 1!`, `Hi 2!` etc.
Inspired by https://glue.tidyverse.org/articles/wrappers.html

teunbrand:
This PR aims to fix #457.

Briefly, a new `label_glue()` function is added that uses the glue package to interpolate strings. The only thing users familiar with the glue package should know is that `x` is the variable that holds the breaks values.

``` r
#> ℹ Loading scales
demo_discrete(iris$Species, labels = label_glue("The Iris {x} flower"))
#> scale_x_discrete(labels = label_glue("The Iris {x} flower"))
```

<sup>Created on 2024-10-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

@thomasp85 are there some steps to take that will make {marquee} integration more useful?


teunbrand:
Added option to parse the labels to expressions:
``` r
devtools::load_all("~/packages/scales//")
#> ℹ Loading scales

demo_continuous(c(0, 10), labels = label_glue("'{number(x)}m'^2", parse = TRUE))
#> scale_x_continuous(labels = label_glue("'{number(x)}m'^2", parse = TRUE))
```



``` r
demo_log10(c(1, 1e4), labels = label_glue("10^{log10(x)}", parse = TRUE))
#> scale_x_log10(labels = label_glue("10^{log10(x)}", parse = TRUE))
```



<sup>Created on 2024-10-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

thomasp85:
in what way do you envision this to interact in a special way with marquee?

I guess using the `marquee_glue` function would insulate it from messing with custom classes but that would require a dependency on marquee. I think we better wait until marquee is ready for the primelight

thomasp85 (approved):
LGTM

PR #460, opened 2024-10-07
fix typos; spelling and missing bracket

kellijohnson-NOAA:
This PR fixes a handful of very minor typos that I found while reading the documentation, only one of which is associated with an issue that is linked in the commit. roxygenize was not executed because there is a GitHub action to do so upon pull requests and I have a different version of the package than the package description.

* continous to continuous
* hypen to hyphen
* adds "to" after "Use `label_number()`
* adds missing ] to [label_date()/[label_time()]


thomasp85 (approved):
Thank you!

PR #461, opened 2024-10-07
New labeller: `label_lut()`

### LINKED ISSUE #458: Feature request: lookup table based label function
The proposal here is to make a new labelling function `label_lut()`, that takes a named vector as input and returns a function that matches the breaks to the names and then returns the matching value.

teunbrand:
This PR aims to fix #458.

Briefly, you pass a named vector and the breaks will be matched to the names, and the value will be returned as the label.

Examples rendered:

``` r
#> ℹ Loading scales

# Example lookup table
lut <- c(
  "4" = "four wheel drive",
  "r" = "rear wheel drive",
  "f" = "front wheel drive"
)

# Typical usage
demo_discrete(c("4", "r", "f"), labels = label_lut(lut))
#> scale_x_discrete(labels = label_lut(lut))
```

``` r
# By default, extra values ('w') will remain as-is
demo_discrete(c("4", "r", "f", "w"), labels = label_lut(lut))
#> scale_x_discrete(labels = label_lut(lut))
```

``` r
# Alternatively, you can relabel extra values
demo_discrete(
  c("4", "r", "f", "w"),
  labels = label_lut(lut, nomatch = "unknown")
)
#> scale_x_discrete(labels = label_lut(lut, nomatch = "unknown"))
```

<sup>Created on 2024-10-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


thomasp85:
Still on the fence about the use of "lut". I don't think many users will immediately understand what it does

Don't know if `label_dict()` or `label_dictionary()` is more apt

thomasp85:
I feel we used the `dict` name somewhere else recently in scales or ggplot2?

teunbrand:
In https://github.com/tidyverse/ggplot2/pull/6077, we'd allow `labs(dictionary = ...)` for a similar lookup.

thomasp85:
then let's mirror that

thomasp85:
also need to rename the test file

teunbrand:
Good catch

PR #463, opened 2024-10-08
`fullseq()` sorts `range`

### LINKED ISSUE #435: `breaks_width()` does not work with reverse transform
To reproduce:

``` r
library(scales)
demo_continuous(
  c(0, 10), 
  transform = "reverse", 
  breaks = breaks_width(1)
)
#> scale_x_continuous(transform = "reverse", breaks = breaks_width(1))
#> Error in seq.default(round_any(range[1], size, floor), round_any(range[2], : wrong sign in 'by' argument
```

<sup>Created on 2024-06-03 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

I think `fullseq()` should sort the range before computing the breaks to fix the issue.
Arguebly, user should use `breaks = breaks_width(-1)`, but that sort of departs from the idea that break function input is specified in original space, not transformed space.

teunbrand:
This PR aims to fix #435.

Briefly, `fullseq()` now sorts the input `range` before doing the calculation. This prevents errors about the sign of the `by` argument.


tests/testthat/test-full-seq.R
```diff
@@ -1,31 +1,36 @@
...
+
   expect_equal(
     fullseq(x, ".5 secs")[1:2],
     as.POSIXct(c("2000-01-01 08:29:58.0 UTC", "2000-01-01 08:29:58.5 UTC"), tz = "UTC")
   )
 })
 
 test_that("works with Date", {
-  x <- as.Date("2012-01-01") + 1:30
-  expect_equal(fullseq(x, "1 month"), as.Date(c("2012-01-01", "2012-02-01")))
```
teunbrand, line 21: This test was wonky. `fullseq()` only considers `range[1]` and `range[2]`, but this is a vector of 30 dates instead of a range of 2 dates.

In addition, `ceiling_date("2012-01-01", "1 month")` gives 1st of march because 31 days from the end of January is in March due to February typically having only 28 days. I consider this out of scope for this PR, but it isn't doing any real harm anyway.

thomasp85 (approved):
LGTM

Issue #443, opened 2024-07-11
Feature request: Could `scales::number` etc have a `style_negative` option of `"none"`?


davidhodge931:
"none" is useful option for population pyramids, where you hack the data by multiplying the Female population 

This should also flow into all of the `label_*()` functions.

So the argument would look like...
```
number(
  ...,
  style_negative = c("hyphen", "minus", "parens", "none"),
```

and the help would have an extra bullet...

* "none" removes any negative styling. 

Could be something for Tidy Dev Day?

davidsirl:
My experience is that this is possible with a slightly different approach to that suggested:
```
scale_x_continuous(breaks = pop_breaks, labels = abs(pop_breaks)),
```
where `pop_breaks` is organised beforehand as `values_plotted |> abs() |> max()` and then using `base::pretty()` to make the breakpoints nice round numbers.

(https://www.njtierney.com/post/2022/08/09/ggplot-pyramid/ was my source for that)

One can "tidy up" such labels further by using comma() but (i) comma() is superseded and (ii) I do not see how to do the same sort of thing with the superseding function label_comma().

teunbrand:
I'm sorry David but I genuinely think this is not a good idea, but I get why it might be useful.
Overall, omitting a minus indication is misleading and we want to make it harder to put up misleading labels, not easier.
I get this then requires some extra effort to label pyramid plots, but this is a specialised plot and the creator should just take some care to `abs()` their values before labelling.

davidhodge931:
Cool, thanks @teunbrand

davidsirl:
I'm not sure I agree  @teunbrand. Any plot can easily be made misleading and, like most plot options, this sort of functionality would aid informative labelling in some situations too. But if that's the decision then fair enough and thank you for explaining it.

PR #465, opened 2024-10-08
Label composer

### LINKED ISSUE #462: Feature request: composable labelling functions
As discussed off-github: it would be nice to have the ability to compose labels by chaining together different functions.

Currently, because all the `label_*()` functions are function factories, it is hard to make small manual adjustments for them and you'd have to build a new function to wrap the operation. Similar to how we have `compose_transform()` we can have `compose_labels()`.

teunbrand:
This PR aim to fix #462.

Briefly, it adds `compose_label()` that you can use to chain together function to format your labels. This saves one the hassle of writing an anonymous function to inject in-between steps.

@davidhodge931 this is a way to solve #443 too :)

``` r
#> ℹ Loading scales
demo_continuous(c(-100, 100), labels = compose_label(abs, label_number()))
#> scale_x_continuous(labels = compose_label(abs, label_number()))
```

<sup>Created on 2024-10-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/label-compose.R
```diff
@@ -0,0 +1,48 @@
...
+  label_list <- lapply(label_list, as_function, call = call)
+
+  function(x) {
+    if (length(x) == 0) {
+      return(character())
+    }
+    orig <- x
+    for (labeller in label_list) {
+      x <- labeller(x)
+      attr(x, "orig") <- orig
```
teunbrand, line 42: Attaching the original breaks as an attribute here so that 'aware' function could use original numeric breaks if they've already been converted to character, or something of the sort.

thomasp85: maybe something more specific than `orig`? `orig_label` is at least more descriptive

teunbrand: I prefer `orig_breaks` then because we attach breaks and not labels as attributes. This confusion just go to show how `orig` was ambiguous 😅

thomasp85:
Thank you

PR #469, opened 2024-10-11
`zero_range()` casts to numeric

### LINKED ISSUE #468: Zero Range for HMS object
The use case for this bug is at https://github.com/tidyverse/ggplot2/issues/6134

The `zero_range` internal function has different behavior for `hms` and all `difftime` objects if there is a "midnight" in there aka `00:00:00` hms value.

# Simple Example

Let's say a function calculated a range of `hms` objects and midnight not in there:

``` r
range = structure(c(1234, 7380), class = "difftime", units = "secs")
scales:::zero_range(range)
#> Error in `/.difftime`((x[1] - x[2]), m): second argument of / cannot be a "difftime" object
```

The error happens at <https://github.com/r-lib/scales/blob/6e5e4c352b36b0903ebcc484264e2e741cd93922/R/bounds.R#L397> and `m` in this case is a `difftime` object.

Create a sequence that crosses midnight, range has a `0` in there, so no error.
The error is “resolved” because <https://github.com/r-lib/scales/blob/6e5e4c352b36b0903ebcc484264e2e741cd93922/R/bounds.R#L391> exits with `FALSE`.

``` r
range_with_zero = structure(c(0, 7380), class = "difftime", units = "secs")
scales:::zero_range(range_with_zero)
#> [1] FALSE
```

teunbrand:
This PR aims to fix #468 and by extension fix https://github.com/tidyverse/ggplot2/issues/6134.

Briefly, we cast input `x` to numeric before doing computations for the range.

``` r
devtools::load_all("~/packages/scales//")
#> ℹ Loading scales

range <- structure(c(1234, 7380), class = "difftime", units = "secs")

zero_range(range)
#> [1] FALSE

ggplot2::resolution(range, zero = FALSE)
#> [1] 6146
```

<sup>Created on 2024-10-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/bounds.R
+++ b/R/bounds.R
@@ -368,7 +368,7 @@ zero_range <- function(x, tol = 1000 * .Machine$double.eps) {
     return(TRUE)
   }
   if (length(x) != 2) cli::cli_abort("{.arg x} must be length 1 or 2")
-  if (any(is.na(x))) {
+  if (anyNA(x)) {
     return(NA)
   }
 
@@ -378,8 +378,8 @@ zero_range <- function(x, tol = 1000 * .Machine$double.eps) {
     return(TRUE)
   }
 
-  # If we reach this, then x must be (-Inf, Inf) or (Inf, -Inf)
-  if (all(is.infinite(x))) {
+  # If we reach this, then x must be (-Inf, <number>) or (<number>, Inf)
+  if (any(is.infinite(x))) {
     return(FALSE)
   }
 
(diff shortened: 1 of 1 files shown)


R/bounds.R
```diff
@@ -378,13 +378,14 @@ zero_range <- function(x, tol = 1000 * .Machine$double.eps) {
     return(TRUE)
   }
 
-  # If we reach this, then x must be (-Inf, Inf) or (Inf, -Inf)
-  if (all(is.infinite(x))) {
+  # If we reach this, then x must be (-Inf, <number>) or (<number>, Inf)
+  if (any(is.infinite(x))) {
```
teunbrand, line 382: Also if just 1 of the 2 numbers are infinite we have non-zero range.

thomasp85 (approved):

PR #470, opened 2024-10-21
Remove unreachable branch of logic in `recycle_common()`

teunbrand:
As pointed out off-github, there was a non-existing `xs` variable that was returned via `recycle_common()`.
Upon further reflection, we can never reach the state wherein that value had to be returned, so in this PR we remove it.


thomasp85 (approved):

thomasp85:
thank you

PR #481, opened 2024-11-19
Train custom classes

### LINKED ISSUE #480: Make sure range training works with custom classes
Only require `range()` and `as.numeric()` to be supported by the custom vector class

teunbrand:
This PR aims to fix #480.

For discrete training, `is.factor(x)` is swapped out by `!is.null(levels(x))`.
That way, any custom class that implements `levels.custom()` should be able to train a range.

For continuous training, we require the result of `range(x)` to be numeric (rather than x itself).
Error message is slightly terser, but that isn't too bad.

``` r
#> ℹ Loading scales
#> Warning: package 'testthat' was built under R version 4.4.2
library(vctrs)

newdisc <- new_list_of(x = list("A", c("B", "C")), ptype = character(), class = "foo")
levels.foo <- function(x) unique(unlist(x))
train_discrete(newdisc)
#> [1] "A" "B" "C"

newcont <- new_list_of(x = list(1, c(5, 3)), ptype = double(), class = "bar")
range.bar <- function(x, ...) range(unlist(x), ...)
train_continuous(newcont)
#> [1] 1 5
```

<sup>Created on 2024-11-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,7 @@
 # scales (development version)
+* Range training now supports custom classes. Continuous classes require a
+  `range()` method that returns numeric values. Discrete classes require a
+  `levels()` method (and optionally a `droplevels()` method) (#480).
 * New `label_glue()` labelling function for interpolated strings (#457).
 * `fullseq()` and by extension `breaks_width()` can now deal with unsorted 
   ranges (#435).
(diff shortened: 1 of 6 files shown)


fabian-s:
this would be awesome and a much needed feature for us -- do you expect this to make it into the next release? is the next release going to come soon?

PR #524, opened 2026-09-17
Resolves #429

### LINKED ISSUE #429: Feature request: could scales support keeping oob values but with infinites turned to NA?
There is currently support of censor and squish with different treatment of Infinite values - but not for keep.

It'd be great if scales could support this.

This could work as a function e.g. `oob_keep_finite`. 

Or alternatively add an infinite argument to `oob_keep`. However, this approach would be inconsistent with the approach taken with the others

math-mcshane:
Uses name proposed by @teunbrand but otherwise functionally behaves as @davidhodge931 suggests. Additionally modifies all `oob` unit tests to include the `NaN` case. Note that `oob_discard` replaces `NaN` with `NA`.


R/bounds.R
```diff
@@ -286,6 +288,14 @@ oob_censor_any <- function(x, range = c(0, 1)) {
   oob_censor(x, range = range, only.finite = FALSE)
 }
 
+#' @rdname oob
+#' @export
+oob_censor_infinite <- function(x) {
+  infinite <- if (TRUE) is.infinite(x) else TRUE
```
teunbrand:
```suggestion
  infinite <- is.infinite(x)
```

math-mcshane:
@teunbrand yes, of course! I was so focused on trying to mimic `oob_censor` that I missed that simplification.

thomasp85 (approved):
LGTM

PR #526, opened 2026-09-17
Add ggplot2 examples to label_*() / breaks_*() docs (#419)

### LINKED ISSUE #419: Add "real" ggplot examples to `label_*()` and `breaks_*()` functions
Currently the examples in the docs for `label_*()` and `breaks_*()` functions use `demo_*()` to showcase what various functions do, but they don't exemplify how to use them in the `scale_*()` layer of a ggplot. It would be good to add a simple, real plotting example as the first example in each of these functions. Then, the demo functions can be left as is to avoid code duplication to showcase various features and for other (testing, etc.) benefits these functions might have.

cynthiahqy:
## Summary

Adds one ggplot2 example to each `label_*()` / `breaks_*()` doc page (21 pages total), as the first example before the existing `demo_*()` examples. Closes #419.

The datasets and plots were planned with Claude per the comments on #419 (now combined into a single comment). Each generated example was then rendered and lightly reviewed by hand for whether the plot actually made sense for the variable being shown — not just whether the code ran — and a few datasets were swapped as a result (e.g. `breaks_exp()` now uses earthquake magnitude/acceleration from `attenu` instead of Old Faithful eruption duration, since there's no reason to exponentially transform the latter). `label_number_auto()` was left without a new example: the generated ones didn't show its effect in a way that was visible in a static plot, and that couldn't be easily fixed within the "one simple example" scope.


thomasp85 (approved):
LGTM

PR #527, opened 2026-09-17
Document available palette names

### LINKED ISSUE #503: Document available palette names
I noticed `palette_names()` was internalized before the release. So, there's no way for users to get the list of available palette, if I understand correctly. I don't think it needs to be exported, but it would be really nice if the sclaes' pkgdown site has a document about the available palettes because it's so many.

jmclawson:
Fixes #503

Some color palettes seem to be named twice, so the first code chunk does some work to figure out which are applicable. Let me know if anything should be changed, and I'm happy to revisit it.


thomasp85:
The pkgdown GitHub action script needs to install quarto - can you add that?

